PR #16, opened 2023-01-22
Readme and installation instruction improvements

ross144:
Built an install script for the open source tool chain for use with Ubuntu 22.04 and 20.04.
Moved book appendix D to README.md as this is likely to change over time.
Added a top level wally block diagram.

There still needs to be some significant cleanup for the install instructions as they reference the book too much.


bin/wally-tool-chain-install.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
```
MikeOpenHWGroup, line 28: Please add an SPDX header.

MikeOpenHWGroup (changes requested):
Thanks @ross144.  Please push in an update to add an SPDX license header on the script.

bin/wally-tool-chain-install.sh
rosethompson: I did not realize the SPDX header needs to be added to scripts as well.  There are number of other scripts which need this change.  I'll add them as well.

MikeOpenHWGroup (approved):

PR #28, opened 2023-01-24
Consolidated SRAMs and removed unused versions

ross144:
- Moved branch predictor files into separate sub-directory.
- Partial BTB cleanup.
- Moved and ranamed btb to btb.sv Fixed btb to use the fixed port 2 sram.
- Removed the old two port ram and replaced it with the fixed version. The fixed version is renamed to ram2p1r1wb.sv
- Removed old versions of gshare.


MikeOpenHWGroup (changes requested):
Can I ask you to add SPDX headers to the modified files?

Just the line starting with "SPDX" is required.

ross144:
I see one file still had the old MIT license. I wonder how that slipped through our original license conversion.  I'll fix it now.

MikeOpenHWGroup (approved):

PR #31, opened 2023-01-27
Modify ram1p1rwbe to add TSMC28nm memories for rv64gc as well as docu…

stineje:
Add docker.md as well as modifications for TSMC28nm memories for cacheway.sv (i.e., generic/mem/ram1p1rwbe)

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,0 +1,107 @@
+# Compliance Development Environment Image
+FROM debian
+
+# UPDATE / UPGRADE
+RUN apt update
+
+# INSTALL
+RUN apt install -y git gawk make texinfo bison flex build-essential python libz-dev libexpat-dev autoconf device-tree-compiler ninja-build libpixman-1-dev build-essential ncurses-base ncurses-bin libncurses5-dev dialog curl wget ftp libgmp-dev python3-pip pkg-config libglib2.0-dev opam  build-essential z3 pkg-config zlib1g-dev verilator cpio bc vim emacs gedit nano
+
+RUN pip3 install chardet==3.0.4
+RUN pip3 install urllib3==1.22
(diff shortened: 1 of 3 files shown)


MikeOpenHWGroup (changes requested):
One "ask" and one "question":

- please add an SPDX license header to all new files.  A two-liner is sufficient, for example:
```
# Copyright 2023 AnAmazingOpenHWContributor
# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```

- The question is about the TSMC memory.  Is this part of the Docker image?  I'm worried about the license implications of even mentioning the module name of the TSMC model in an OpenHW repo.

davidharrishmc:
James,

What if we replace the TSMC module names with a generic module name that refers to a wrapper outside the OpenHW repo?

David

MikeOpenHWGroup:
Would this mean that the Wally core in the CVW project could not be compiled without a TSMC memory model?  If that is the case, can we create a generic RTL-simulation-only memory model?

davidharrishmc:
We have a configuration parameter USE_SRAM.  It defaults to 0, which describes memories with always @(posedge clk), suitable for simulation and basic synthesis.

We are now incorporating 28 nm memories for an improved synthesis.  We set USE_SRAM to 1 in this case. 

The issue is whether to include the name of a TSMC memory in the Wally repo, or just to name a generic wrapper that is defined outside the Wally repo alongside the TSMC memory.  I think we are leaning toward the second option.

David

stineje:
Mike,

I hope to push this in about 5m.  I will remove any word of the TSMC name.  Take care.

All my best,

James

MikeOpenHWGroup:
Thanks for the clarification.  All good!

PR #36, opened 2023-01-29
RV32imc configuration

davidharrishmc:


ross144 (approved):
I approve. I don't see anything which would cause issues.

ross144 (approved):
Alignment is a bit messed up in several files when replacing `XLEN with XLEN, etc.  Since we are likely to mess up alignment further lets accept for now with the plan to fix once the things settle down a bit.

PR #41, opened 2023-01-31
Merged imperas branch into main. Remove old branch when pull request accepted.

ross144:
- Fixed issue with rvvi tracer so it reports call csr changes, not just instrutions which write the CSRs.
- Found a potential issue with mstatush when XLEN = 64.
- Code refactor and addition of rvvi interface
- refactor all rvvi into single initial block
- Fixed the rvvi CSR write enable not synchronized with a valid instruction in the Writeback stage.
- update for private copy of Imperas
- ignore external
- refer to correct path
- remove volatile for FFLAGS and FCSR
- add im flags for compressed disass
- changes made with Ross
- Partial fix for misaligned LD/ST
- Modified to clone imperas via git rather than https.
- customer commands
- correct the HASH
- update
- Imperas found a bug with the Fence.I instruction. If a fence.i directly followed a store miss, the d$ would release Stall during the cache line write. Then transition to ReadHold.  This cause the d$ flush to go high while in ReadHold.  The solution is to ensure the cache continues to assert Stall while in WriteLine state.
- Minor gshare optimization.
- Simplified gshare.
- Major cleanup of branch predictor.
- Improved signal names.
- More branch predictor cleanup.
- Parameterized testbench branch predictor preload.
- Fixed parameterization in testbench.
- Found small bug in gshare.
- Renamed signals in RAS.
- Minor bug fix in gshare.
- Fixed remaining bugs in the imperas merge.


MikeOpenHWGroup:
Go Cowboys?!?

stineje:
Wrote this when I joined Oklahoma State - we are the Oklahoma State Cowboys.  We probably never removed.

MikeOpenHWGroup:
> Wrote this when I joined Oklahoma State - we are the Oklahoma State Cowboys.

Yeah, I figured that.  Couldn't resist teasing you about it.  Also: yes, I read the content of pull-requests :wink:

PR #63, opened 2023-02-04
Cleanup

davidharrishmc:
Simplified names of do file
Fixed licenses
working on debug.S, but it doesn't yet run in Spike.  Illegal instruction error on floating-point code


davidharrishmc:
Pushed because Madeleine needs it.

PR #81, opened 2023-02-15
Fix error in else in ram1p1rwbe that causes synthesis to miss things.…

stineje:
* Fix wrapper in ram1p1rwbe to avoid issue with else
* Put back wrapper that got inadvertently changed


stineje (commented):
Update

ross144:
This may not have been the thing I wanted to do.  James wants to remove his pull request before fixing the issue.

PR #92, opened 2023-02-18
Update status in synthesis for physical synthesis

stineje:
Update memories and some synthesis items for psyn


davidharrishmc:
Declined PR.
Avoiding using ROMs.
Avoid using $config in dc_setup

PR #98, opened 2023-02-20
New gshare implementation

ross144:
- keep this commit off of cvw.
- dc shell setup.
- Fixed bug with branch predictor.
- Fixed a branch predictor performance issue.
- Updated global history predictor.
- More updates.
- Fixed global history predictor.
- Renamed globalhistory predictor.
- Minor fix.
- Possibly much better branch predictor implemention. The complexity is significantly reduced.
- Simplified branch predictor.


davidharrishmc:
Ross, commit 27f6 is called Keep this Commit of CVW.  Should it be committed?  It seems to have a fair amount of redundancy with commits I've already made.

ross144:
It does look like significant overlap with your commits.  I think my merge 0d79c0c resolved the conflicts between yours and mine.  It would be worth taking a closer look however.

PR #99, opened 2023-02-20
Add setup.csh for C shell users at OSU

stineje:
Add setup.csh for OSU users


ross144:
We need to add the license.

PR #100, opened 2023-02-20
Synthesis improvements

davidharrishmc:


MikeOpenHWGroup:
@davidharrishmc, @ross144, please do not merge in PRs with failing ECA checks.  If you are having troubles with this, reach out to either me or @DBees.

ross144:
It's not clear to me why it's failing the ECA check.  Commits a59526f and da6064e have the same author and email address.  I thought we put in a request about this last week but we had not heard anything back about what is causing the issue.

MikeOpenHWGroup:
Yeah, I hasven't heard back beyond, "it was an intermittent". Now that this one is merged, I cannot check it (feature of the EF interface). If you see it again let me know.

ross144:
Is there a work around for the time being?  We need to be able to continue to merge changes in the mean time.

MikeOpenHWGroup:
AFAIK, the issue was intermittent and "went away".  Is that not the case?  If you have something quick to merge we can try it now.

MikeOpenHWGroup:
ECA check for #101 worked.

ross144:
Hmm. Looking though the past pull requests.  Typically only the last commit is the one to not pass the ECA check.  However all the commits in the merge are labeled.
From: David Harris <david_harris@hmc.edu>
I wonder if the system just reports the last commit as wrong when multiple are not passing.  (There is a counter example in this PR.)  How do we check which email address @davidharrishmc has on record?

ross144:
Looks like my theory is wrong.  Some older merges used the same email david_harris@hmc.edu but did not have the ECA failure.

davidharrishmc:
The problem isn’t intermittent.  It’s been showing up on all of my commits in the last few days.  I’m not aware of changing my email on either git or the ECA.  

However, I went to the ECA agreement and found my email was listed as ***@***.***  I’ve changed it in the ECA to ***@***.*** to match the
git commits.  After a whole bunch of clicking and retrying and resubmitting, I finally seem to have it all in sync and the PR 104 was validated.

I didn’t have problems last month and am unsure what changed to trigger this, but it seems to be solved now.

stineje:
Mike,

This happened to me where my system similarly to what David encountered.  Unfortunately, I was not an approver, so I had to get my submission invalidated and resubmit.  I did this through a git config.  I still think there is a problem on OopenHW’s side about the ECA.  In the past, I have done a merge with no problems.  Therefore, I am not sure what the issue is and it may be very difficult to debug.  David and Ross are always diligent on their side so I find it odd they had problems too.  Take care.

All my best,

James

MikeOpenHWGroup:
Let's take this to MM.  I've set up a channel for us and roped in @DBees as well.

PR #102, opened 2023-02-20
Script updates

davidharrishmc:
Let's also test if there are ECA problems here


davidharrishmc:
Please test this PR#102.  It is showing an x on commit 04bb2 and no x on commit 023ba6 even though both come from david_harris@xxx.edu.

PR #110, opened 2023-02-23
Added support for bit manipulation extension

kipmacsaigoren:
Support has been added for the bit manipulation extension specified [here](https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions). 
CVW should now work with Zba, Zbb, Zbc, and Zbs extensions. 
In addition, tests and makefiles have been updated to correctly compile and simulate the arch tests for these extensions in rv32 and rv64. 
All other tests and configurations still pass whether or not these extensions are enabled.


kipmacsaigoren:
I have signed up for an Eclipse foundation account using my primary github email, kmacsaigoren@hmc.edu. I've also signed the ECA with the same email and included my github username in my eclipse account. The check still won't go through. are there any recommendations for me to pass this requirement?

davidharrishmc:
This is a big one.  Let's review together before I pull it into the main branch.

PR #129, opened 2023-03-06
Further illegal instruction detection

davidharrishmc:


ross144:
Looks like PR #128 broke regression and this one fixes the problem.

PR #138, opened 2023-03-13
Fix MISA RO and UART addresses

eroom1966:
It appears on inspection that the MISA register is read only in Wally In which case this has now also been set in the ImperasDV representation Also the Addresss for the UART R/W privileges are corrected

@ross144 @davidharrishmc 
Could you please confirm that MISA is Read only, otherwise this will need correcting in the tool configuration file


ross144:
Yes the MISA is read only.

davidharrishmc:
Confirmed

PR #176, opened 2023-03-30
Fixed regression after cause CSR fix

kipmacsaigoren:
Reducing the number of bits in the cause and edeleg CRSs caused an issue with the wally priv tests.
This pull fixes it and updates the writing of both CSRs to match the Imperas DV tests.

Requires a make allclean


ross144:
Can you clarify why the extra bits are necessary?

PR #178, opened 2023-03-30
Improve I$ coverage by simplifying logic

AlecVercruysse:
Hi all, this PR hopes to improve I$ coverage as part of Lim and I's E154 final project.

@ross144, Prof. Harris has reviewed everything else but was hoping you might be able to take a look at the proposed change to the LRUMemory reset. We've written up a bit about the change in this project report [google doc](https://docs.google.com/document/d/15kmg0c1-cOloDY3EaORa20bwu9_kD3s5c1ZRB6zqN0E/edit) which you should have access to.

A summary of proposed changes:
 - Make the cache write path conditional on whether or not it is read-only
 - Remove the LRUMux that corresponds to the root (always hit) node of the LRU tree.
 - Remove an unhit case of the LRUMemory reset logic (unreviewed!)

Regression passed on everything except for timing out on priv and periph, but those cases timed out even on the main branch for me before adding the changes (I'm a few commits behind upstream).

Thanks!


ross144:
Hi Alec and Lim.  This all looks good.  My apologizes for taking so long to reply I was dealing with a personal matter for the last four days.

PR #179, opened 2023-03-30
Fixed broken regression: privileged tests and build root

davidharrishmc:


MikeOpenHWGroup (approved):
Approving/merging as per @davidharrishmc request as the other Mr. Thompson (@ross144) is unavilable.

PR #190, opened 2023-04-03
expanded ifu coverage including 4 added directed tests and 1 exclusion, expanded fpu coverage including 6 directed tests and 2 multiline exclusions

SydRiley:
4 files were edited

1) src/fpu/fctrl.sv
- Added 2 multiline coverage exclusions to exclude branches that cannot be covered by the rv64gc configuration because it doesn't support floating point half or quad precision

2) src/ifu/decompress.sv
- Added 1 exclusion of a branch that will never be hit in the rv64gc configuration. The branch requires `XLEN !> 32 and in rv64gc, `XLEN is 64.

3) tests/coverage/ifu.S
- Added 4 tests to cover 4 previously uncovered branches/instructions (c.fsd, c.fldsp, c.fsdsp, an illegal instruction)

4) tests/coverage/fpu.S
- Added 6 directed tests to cover All False branch for 6 different sets of cases. 

The ifu and fpu now have full branch coverage. My next step will be improving the conditional coverage.


davidharrishmc:
Nice work, Syd. Cleanly implemented and well documented.

PR #199, opened 2023-04-04
Fixed WFI to commit when an interrupt occurs

davidharrishmc:


davidharrishmc:
Fixed 174, 183, 184 in conjunction with PR198.

Still failing WALLY-wfi-01 privileged test; others passing.

PR #206, opened 2023-04-05
i$ coverage improvements

AlecVercruysse:
A lot of I$ coverage issues are due to the fact that the read-only cache and rw-caches are described with the same verilog code. There were many opportunities to make the cache dirty/flush/write logic somehow dependent on READ_ONLY_CACHE to improve coverage.

Another significant change is adding ram1p1rwe, which does not have byte-enables. This is useful for the data memory of read-only cache, as well as the tag array for any caches, both of which always have every byte-enable high. One thing to note, however, is that coverage is disabled for the chip-enable during a write of this new memory.

I've written my reasoning for each change in the commit message for the change (and attempted to separate commits as much as possible). One extra thing to note are that for commit `3867142`, the cachetagmem was also changed to ram1p1rwe for the D$, not just the I$. This makes sense since the byte-enable is not needed for tag memory. Turning off coverage for the chip-enable in ram1p1rwe, however, might adversely exclude coverage for the D$ cachetagmem, since it might be possible that a write could when the chip-enable is pulled low for D$. I would appreciate suggestions for another way to deal with the chip-enable coverage for i$ memories if this is unacceptable.

Thanks!

### DIFF
--- a/src/cache/cache.sv
+++ b/src/cache/cache.sv
@@ -76,7 +76,7 @@ module cache #(parameter LINELEN,  NUMLINES,  NUMWAYS, LOGBWPL, WORDLEN, MUXINTE
   logic [1:0]                    AdrSelMuxSel;
   logic [SETLEN-1:0]             CacheSet;
   logic [LINELEN-1:0]            LineWriteData;
-  logic                          ClearValid, ClearDirty, SetDirty, SetValid;
+  logic                          ClearDirty, SetDirty, SetValid;
   logic [LINELEN-1:0]            ReadDataLineWay [NUMWAYS-1:0];
   logic [NUMWAYS-1:0]            HitWay, ValidWay;
   logic                          CacheHit;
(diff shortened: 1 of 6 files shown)


ross144:
While I understand the drive to use a minimized ram1p1rwe for the read-only cache, ASIC and FPGA SRAM memories are nearly always have byte or bit write enables.  I believe it is better to model the behavior of these memories accurately rather than push for higher code coverage.  I suspect there is a way to wavier these impossible to reach states.  I am open to having my mind changed but at this time I am leaving this pull request open.

AlecVercruysse:
Hi Ross, I spoke with Prof. Harris today, and he had the opinion that as long as the ram1p1rwe module instantiated the same underlying SRAM macros (and tied the byte enables to active), the ram1p1rwe module was the least bad solution to this gap in coverage. 

The problem with adding a waiver is that the coverage issues appear in the ram1p1rwbe module specifically, and waiving any coverage of a byte enable in there might hide coverage issues for other instantiations of the module (e.g. in the D$, where we do want to try different byte-enable coverages).

ross144:
I can be amenable to that solution for ASICs.  I wonder how this will work for the FPGA block ram.  Let's merge it for now and see if it causes an issue.  If it causes a problem we'll revisit.

ross144:
Some good news and bad news. This PR did not break the functionality of the FPGA linux boot, but it does implement the i$ as disturbed RAM instead of block ram.  I'm not sure how I want to proceed at this point.

davidharrishmc:
Interesting.  What is it about the coding that prevented block RAM?  Is there a tweak to get around this?

ross144:
We could modify the generate around the ram1p1rwbe and ram1p1rwe to select ram1p1rwbe if either FPGA or not READ_ONLY_CACHE.  Seems a little hacky though.

davidharrishmc:
Does this have anything useful?

I’ve inferred block RAM years ago without a byte enable.  I’m still suspicious that it’s something else minor.

David

https://support.xilinx.com/s/article/46515?language=en_US￼

ross144:
Hmm. There guide shows inferring block ram without the byte write enables. I did a quick experiment to further simply the logic and it still uses disturbed ram.

davidharrishmc:
Alec,

I’ll leave this to you and Ross to decide.  If you two don’t have a viable path toward block RAM synthesis of the memory without byte enables, it sounds like you may need to do this.

In any case, this looks like a solution to Syd’s problem of excluding certain unused address decoders.

David

PR #207, opened 2023-04-05
Cache Simulator

magpyed:
This adds a L1 cache simulator to cvw/bin. This simulator can be invoked as follows:
CacheSim.py [number of lines] [number of ways] [physical address length] [tag length] -f [log file].
An example invocation for rv64gc would be CacheSim.py 64 4 56 44 -f [log file].
These log files are generated via the I/D$ logger in testbench.sv, which I have also altered. The log format is now 'address [R/W/A/I/F] [H/M/E/D],' with the second entry being access type (read, write, atomic, invalidate, flush), and the third being result (hit, miss, eviction, dirty eviction/writeback). The timing of the log writes has been altered - each line is now written when LRUWriteEn is high, allowing us to catch evictions. This timing could likely use more tweaking in the future.
To try and confirm the correctness of this simulator, a test has been added in cvw/tests/custom/cacheSimTest. It uses a toy-sized address to test the basic functionality of CacheSim.py, including the pLRU tree. This folder also contains reference simulator output for the two rv64gc test suites that are failing: arch64i and wally64priv.


ross144:
Do we need to include the *.out files in the commit?  Can they be excluded?

magpyed:
Ah, yeah, they can be excluded - my bad!

PR #214, opened 2023-04-06
Add in configuration for B extension

eroom1966:
this will allow the execution of bit manipulation in the Imperas DV environment


davidharrishmc:
Thanks.  Looks like we overwrote each other on bitmanip, but now it is good again.

PR #220, opened 2023-04-08
Obscure coverage fixes

davidharrishmc:


infinitymdm:
Good to see that StallM issue get figured out! I had been working on that one for a little while. Looks like perhaps I took the wrong approach, but I'm glad to see it resolved.

PR #221, opened 2023-04-09
Added test coverage for Privilege Unit in CSRs

dherreravicioso:
Hi all,

These changes add test coverage for floating point registers, some PMP addresses that were uncovered, and the MTVAL, MCAUSE, and MCOUNTINHIBIT CSRs. All changes pass regression. 

Thanks!


davidharrishmc:
Whoo!

PR #222, opened 2023-04-09
Remove unnecessary check from compressed instruction decode

kjprime:
The following changes remove a unreachable state in coverage within the decompress file and redundant check. 
This statement is checking for a value aligning to the last possible option in all cases for truth table for instruction bits 12:10.

### **Truth table instr16[12:10]**
``` 
line#  | 12 | 11 | 10
114    |  X   0    0
116    |  X   0    1
118    |  X   1    0
120    |  0   1    1
129    |  1   1    1
```
I believe checking this statement is redundant and leaving it as a comment increases clarity.


davidharrishmc:
Whoo!

PR #228, opened 2023-04-10
Deleted the ~BadNaNBox section in the subnormal FP equation

Dygore:
After removing this section of the equation, the 4 subnormal test case errors were resolved making the fclassify section of the FPU fully covered!


davidharrishmc:
Dylan, I don't think this is correct.  When you delete some code to seek coverage, you need to justify why it was wrong.

In particular, I think the code was needed for an incorrectly NAN-boxed subnormal.  For example, if f1 = 0xF0FFFFFF_000000001 and we do an fclass.s, it should look at the bottom half and find that it is subnormal.  However, the upper half should be all Fs to be NAN-boxed, and it isn't, so instead the answer should be a NaN.

Do you know how to add a code sequence to test this?  It should exercise the line and get coverage.

davidharrishmc:
@Dygore it would be good to respond to feedback on your PR in a timely way.  I'll close this PR unless I hear otherwise from you shortly.

PR #233, opened 2023-04-12
Full I$ coverage 

AlecVercruysse:
This PR should bring the i$ coverage up to 100%.

Big changes include
 - a big refactor of the cachefsm.
 - Excluding from coverage the potential case where `FlushStage` (`FlushD`) asserts simultaneously as a write (`SetValidWay`) for the read-only cache. My reasoning is in the commit message (3fc6bb0c), but in short, I think it's impossible to hit.
 - Changing some adress Muxes from mux3 to mux2 for read-only caches.
 - Fixing last LRUUpdate coverage issues, and changing some genvar names to improve readability. Note that the textbook's description of LRU has some outdated signal names as well. 

Note that the rv64gc_wally64priv test failed for me, but it failed for me on [2f6ed64](https://github.com/openhwgroup/cvw/commit/2f6ed64e26ef2940f1627c7e9266f04f158b0303) (the current upstream/main) as well. Please let me know if this is something unique to me!

### DIFF
--- a/src/cache/cacheway.sv
+++ b/src/cache/cacheway.sv
@@ -97,9 +97,9 @@ module cacheway #(parameter NUMLINES=512, LINELEN = 256, TAGLEN = 26,
   /////////////////////////////////////////////////////////////////////////////////////////////
 
   assign SetValidWay = SetValid & SelData;
-  assign ClearDirtyWay = ClearDirty & SelData;
   if (!READ_ONLY_CACHE) begin
     assign SetDirtyWay = SetDirty & SelData;
+    assign ClearDirtyWay = ClearDirty & SelData;
     assign SelectedWriteWordEn = (SetValidWay | SetDirtyWay) & ~FlushStage;
     assign SetValidEN = SetValidWay & ~FlushStage;
   end
--- a/src/cache/cacheway.sv
+++ b/src/cache/cacheway.sv
@@ -97,22 +97,10 @@ module cacheway #(parameter NUMLINES=512, LINELEN = 256, TAGLEN = 26,
   /////////////////////////////////////////////////////////////////////////////////////////////
 
   assign SetValidWay = SetValid & SelData;
-  if (!READ_ONLY_CACHE) begin
-    assign SetDirtyWay = SetDirty & SelData;
-    assign ClearDirtyWay = ClearDirty & SelData;
-    assign SelectedWriteWordEn = (SetValidWay | SetDirtyWay) & ~FlushStage;
-    assign SetValidEN = SetValidWay & ~FlushStage;
-  end
-  else begin
-    // Don't cover FlushStage assertion during SetValidWay.
-    // it's not explicitely gated anywhere, but for read-only caches,
-    // there's no way that a FlushD can happen during the write stage
-    // of a fetch.
-    // coverage off -item e 1 -fecexprrow 4
-    assign SelectedWriteWordEn = SetValidWay & ~FlushStage;
-    // coverage off -item e 1 -fecexprrow 4
-    assign SetValidEN = SetValidWay & ~FlushStage;
-  end
+  assign SetDirtyWay = SetDirty & SelData;                                 // exclusion-tag: icache SetDirtyWay
+  assign ClearDirtyWay = ClearDirty & SelData;
+  assign SelectedWriteWordEn = (SetValidWay | SetDirtyWay) & ~FlushStage;  // exclusion-tag: icache SelectedWiteWordEn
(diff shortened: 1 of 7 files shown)


magpyed:
Hey Alec! I would personally undo the FlushStage changes to the cacheLRU - while LRUWriteEn is definitely always used alongside FlushStage, I personally feel that making cache.LRUWriteEn not equal to cache.cacheLRU.LRUWriteEn will create confusion in the future. So, I'd leave FlushStage as an input, and restore the "& ~FlushStage"-es in the LRU. This is my personal feeling, though, and I'd be happy to hear other opinions.

ross144:
I like the name changes for the localparams in cacheLRU. 

I'm not sure I understand why we need to factor out FlushStage from CacheLRU.  I prefer the previous style as it matches whats used through the rest of Wally.   I get in this case the critical paths will be the same, but the old style matches the rest of Wally.  If you have a convincing argument I'm willing to allow this change; otherwise I prefer revert back this one.

I'm also not sure how I feel about the duplication of the fsm into read and write caches.  This has a major downside of reducing code reuse.  I'd strongly prefer a different approach than making two separate fsms.  The synthesizer is smart enough to simplify the logic so I'm not concerned about extra hardware.  I'd like to explore coverage exclusions before committing to this type of change.  My biggest concern is if we find a bug in the cache fsm we have to patch it in both implementations and this becomes a maintenance problem in the future.

I know I'm being a stickler here. 

TL;DR. I want to find a solution which does not require us to refactor the code into two separate FSMs but still gets us coverage.

AlecVercruysse:
Point taken about how the difference between LRUWriteEn in cache and LRUWriteEn in cacheLRU might cause confusion. However, LRUWriteEn is completely internal to the cache, generated by cachefsm. I don't see why both FlushStage and LRUWriteEn both need to be passed to CacheLRU in that case, especially if doing so results in missed coverage.
Furthermore, I would argue that in clean code, the number of connections between modules should be minimized, and if LRUWruteEn is /always/ gated by ~FlushStage, why not include that in the definition of the signal? Does it make sense to change the assignment of LRUWriteEn in the cachefsm module to include `& ~FlushStage)` ?

As for the code duplication in my changes to cachefsm: I will take a look at writing some I$-specific exclusions in the .do exclusion file. We've discovered you can exclude by scope in some cases so there probably is a better way.

magpyed:
I like changing the LRUWriteEn assign statement in cacheFSM.

AlecVercruysse:
Okay, I think I have a method of excluding specific logic in just the I$ instance of cachefsm that won't break immediately when edits are made.

Basically, it all goes in the exclusion .do file, and all exclusions use the `-scope` flag to limit their applicability to just I$. We still need to specify a line number, however, to drill down to specific branches/conditions/expressions/etc to exclude.

I have a tcl procedure in `GetLineNum.do` that searches for a regex in a source file and returns the line number. So in the source code, I just write something along the lines of a self-defined pragma: `// exclusion-tag: icache CacheW`, and then in the .do file I can to a regex search for that tag to find that specific line number. This way, the exclusion stays attached to the code, and not the line number.

The only issue is that in many cases, I exclude only specific lines of the FEC expression table. I believe the order of the rows in that expression table depends on the order in which the signal names appear in that line of code. So if that line of code is ever edited, the exclusion probably needs to be revisited as well. On the bright side, having an `// exclusion-tag` present at that line of code clues the reader in that there is an exclusion to be aware of. 

I believe this is the least-bad solution to the i$ fsm coverage issue.

ross144:
This sounds like a great solution.  Editing the exclusion code after an update is an unfortunate cost, but like I like the exclusion-tag comment idea.  Great idea.

src/cache/cacheway.sv
AlecVercruysse, line 1: This is logic that I had changed in a previous commit that I reverted. I checked the diff against an old branch to ensure the logic is the exact same. (It also passes regression of course).

AlecVercruysse:
@davidharrishmc I think the PR is ready to go.

With this new coverage logic, I think we can replace the ram1p1rwe that I introduced with the byte-enabled version, to fix the issues of block ram not getting inferred on FPGA synthesis. @ross144 I can do this in my PR for next week if the issue still persists. The real shame here is that the -scope argument of exclusion commands doesn't accept wildcards, so the number of cacheways is hard-coded in the exclusion file. This hardcoding of cache parameters would be exacerbated (i.e. other parameters would also need hardcoding)  by switching back to the byte-enabled ram.

ross144:
This looks great!

PR #241, opened 2023-04-13
Added a test for denormalized FP numbers

Dygore:
I added a test in fpu.S to add coverage for the denormalized numbers in the fclassify.sv file. This test was created in reference to Ross's guidance and was tested to ensure proper compiling and coverage.

I made a mistake in thinking the test couldn't be reached previously however, as is evident by the test it actually can be reached it just wasn't fully covered in the test bench. This test erased the remaining coverage lapses for the fclassify section of the Wally processor.


davidharrishmc:
Nice work!  Good to get this one solved!

PR #244, opened 2023-04-14
Added tests for full coverage of the FPU result sign module

Dygore:
Added two tests to the fpu.s test file to get full coverage for the result sign module.

First I added a test with an infinity FP value as one of the addends which added coverage for most of the unit.

Then I added a test to add coverage for the fmaAs for Z sign bit on the addend of the FMA which I made subtract to make the bit 1. 

These tests took the result sign module from 71.19% covered to 100% covered!


davidharrishmc:
@Dygore I accepted your changes, but they don't actually cover anything new.  I think you were running regression-wally with the arch64d and f tests commented out.  Those tests fully cover resultsign already.  Be sure to restore those tests so you have a full view of the FPU coverage.  the modules that still need coverage are:

unpackinput
fdivsqrt (I'm working on that already)
cvtshiftcalc
round
flags
specialcase

Dygore:
@davidharrishmc Thank you, I didn't realize those were already covered, I'll shift my attention to the the modules not already covered!

PR #248, opened 2023-04-16
Added test coverage for reads to HPM counters and coverage exclusions

dherreravicioso:
Hi all,

This PR contains coverage exclusions for:
1. Misaligned instructions, as they are impossible in rv64gc
2. MSTATUSH, as that register will only be used in 32-bit configurations 

In addition, we added some tests to write to HPM counters from different privileged modes with different conditions of the SCOUNTEREN and MCOUNTEREN registers. 

Thanks!


davidharrishmc:
Nice work.  Nice figuring out how to exclude only one term from the coverage off expression!

PR #254, opened 2023-04-18
Added coverage test for fcmp.sv

cocobyte1:
Added test to fpu.S for line 140 of fcmp.sv to hit (YNaN)_1.


davidharrishmc:
I see this file indicates coverage for fcmp.sv.  

fcmp.sv already has 100% coverage if you are running all of the tests.  Did you remember to adjust regression-wally to run the arch64f and arch64d tests?

PR #263, opened 2023-04-20
Add files via upload

BigSwoll01:
Added fmv.x.h and fmv.h.x testing for fpu.S

### DIFF
--- a/fpu.S
+++ b/fpu.S
@@ -1,0 +1,156 @@
+///////////////////////////////////////////
+// fpu.S
+//
+// Written: David_Harris@hmc.edu 28 March 2023
+//
+// Purpose: Test coverage for FPU
+//
+// A component of the CORE-V-WALLY configurable RISC-V project.
+// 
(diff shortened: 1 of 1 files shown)


BigSwoll01:
Let me know if there is something else for me to do here

DBees:
@BigSwoll01 I'm not sure why this validation check is failing. I just checked your eclipse account and it seems ok...

davidharrishmc:
I'm not sure why this request is failing either.  However, the PR is trying to replace the entire file, which isn't how a commit is supposed to be used.  

I'd suggest cancel this pull request and discard your changes to the fpu.S file.  Make sure you've pulled the latest version of the repository.  Reenter your specific changes and commit and push them.  Try making a PR again and see if anything different happens.  At least we will know which lines you are trying to change.

Also, please indicate your coverage before and after the change and which module you are covering with your tests.  Give enough detail that the rest of us can repeat the coverage sim and verify that we get the same benefit you saw.

PR #267, opened 2023-04-21
Update fpu.s to add coverage for negative denormalized numbers

Dygore:
I added a test for negative denormalized numbers in the FPU.s

This increased the coverage of fclassify from 85% to 100%

It's similar to a previous PR I did, I just added a test for the sign bit being negative and denorm to be true.


davidharrishmc:
@Dygore I think there's been a miscommunication.  Run ./regression-wally -coverage -fpu to include all of the FPU tests.  You'll find fclassify already had 100% coverage when the full set of tests are used.

PR #277, opened 2023-04-25
fmv.x.h and fmv.h.x Testing added

BigSwoll01:
We were told to start small so I added fmv.h.x and fmv.x.h to the fpu.S

No other tests were changed, thus nothing else was harmed in the process of increasing coverage.

Please let me know if i need to do anything else.

### DIFF
--- a/tests/coverage/fpu.S
+++ b/tests/coverage/fpu.S
@@ -36,6 +36,12 @@ main:
     flw ft0, 0(t0)
     fclass.s t1, ft0
 
+ # fmv.x.h and fmv.h.x Testing
+    la t1, TestData1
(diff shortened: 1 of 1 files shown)


BigSwoll01:
does not work asexpected

infinitymdm:
FYI, you should typically test your changes and make sure they work as expected before you submit a Pull Request.

A PR notifies the upstream that you've made improvements and are ready for them to be shipped out to other users of the software. If you aren't confident that your code is ready for users, you shouldn't make a PR.

BigSwoll01:
I will make sure of that for next time. thanks

PR #279, opened 2023-04-26
Excluded and added coverage for WFI test case.

ACWright256:
Removed coverage for a test case involving the WFI. This case involved testing a case where the WFI is not supposed to trap, therefore causing a watchdog timeout. This PR also includes test cases for a similar case where it is supposed to trap. This brings priv coverage from 99.75% to 100%!


davidharrishmc:
Could you comment the exclusion a bit further so the reader will understand exactly which case is being excluded and why?

davidharrishmc:
Whoo 100% for privileged!

PR #329, opened 2023-06-12
Fixed the spacing for the generic, ieu, ifu, lsu, mdu, mmu, and privileged modules

harshinisrinath1001:


src/ifu/CodeAligner.py
davidharrishmc, line 1: CodeAligner.py doesn't belong in the src/ifu directory.  
I'd suggest keeping it local for you but not putting it in the repository.

davidharrishmc (changes requested):
Please remove CodeAligner.py from the PR.

PR #333, opened 2023-06-14
cvw.sv moved to root to avoid warnings; UART cleanup and QEMU removal

davidharrishmc:


ross144:
Hmm. I'm a bit hesitant to completely remove the qemu config option yet.  Should we try linux with imperasDV first?

davidharrishmc:
Take a look at the UART changes.  I think you'll find they are innocuous.  It's not removing QEMU from the whole system, and should remain compatible with QEMU simulation.  But it's worth getting your judgment.

PR #351, opened 2023-06-28
fp-testbench does not stop on error now

kevindkim723:


stineje (commented):
looks like removal of $stop - wouldn't it be better to $stop on an error.  Perhaps, we can add a feature to testfloat.do to check and stop if needed.  I can make this change in about 60sec if you want this.

ross144:
I like that idea.

stineje:
Made new PR to help avoid this issue - kevindkim723: if this is okay, you can cancel this PR.

davidharrishmc:
Kevin, just comment out the stop in your local version while you need to run past first error.  Let's leave the stop in the repository.

PR #353, opened 2023-06-29
Add flag in testfloat.do to avoid stopping on error for Kevin

stineje:
Add option to avoid stopping on error for Kevin


davidharrishmc:
I’m reluctant to add this complexity to code in the repo.  It feels like a hack and it won’t be needed when the test cases pass.  What is failing, and can Kevin just comment out the stop temporarily if he needs to run past first failure?

stineje:
Yes, he can just comment out which is what I recommend too - should be cool.

PR #355, opened 2023-07-02
Decoder improvements

davidharrishmc:
Significantly improved detection of illegal instructions, including unsupported Cache Management Operations instructions (per Imperas bug report)

Added initial support for CMOE coming out of the IEU.  Ross can use these to implement CMO.  Cache needs to look at Instr[22:20] to determine which type of CMO instruction.

### DIFF
--- a/src/fpu/fctrl.sv
+++ b/src/fpu/fctrl.sv
@@ -101,13 +101,13 @@ module fctrl import cvw::*;  #(parameter cvw_t P) (
       /* verilator lint_off CASEINCOMPLETE */   // default value above has priority so no other default needed
       case(OpD)
         7'b0000111: case(Funct3D)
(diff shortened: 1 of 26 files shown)


ross144:
@davidharrishmc Can we change the comment on line 100 of controller.sv to include the additional bit?

davidharrishmc:
I'm not sure what comment you are talking about.  Line 100 for me is

  logic [1:0]  MemRWD, MemRWE;                 // Store (write to memory)

The key comment at 231 has the extra CMO field labeled. Is that what you were looking for?  Otherwise, feel free to add the comment I'm overlooking.

    // RegWrite_ImmSrc_ALUSrc_MemRW_ResultSrc_Branch_ALUOp_Jump_ALUResultSrc_W64_CSRRead_Privileged_Fence_MDU_Atomic_CMO_Illegal

ross144:
Sorry. I had the wrong line number. That's the correct line 231.  I'll go ahead and merge.

PR #365, opened 2023-07-21
Boot

JacobPease:
I've updated the boot process to use OpenSBI and Linux binaries. These binaries are now stored in their own GPT partitions on the SD card and loaded into memory at the correct addresses, specified by the 'generic' configuration of OpenSBI.


ross144:
@MikeOpenHWGroup Hi Mike.  I am wondering why Jacob's pull request is failing the ECA check.   He signed the ECA with his jacobpease@...com email which match all his commits.

ross144:
Looks like Jacob's account is ok now, but there are anonymous user commits.  These looks like old commits from Dr. Stine.  Jacob's pull request contains a bunch of merge requests from past commits so I think these are already in our repo.  I'm going to spot check a few and if they are already in there I'm going to go ahead an approve this pull request.

PR #377, opened 2023-08-03
Update combined_IF_vectors scripts to check if the WALLY environmenta…

stineje:
If the wally environmental variable is not set, this script will not work.  This tests for this environmental variable which may occur for those working just on floating-point stuff.


stineje:
mistake - will fix

PR #400, opened 2023-08-28
configurability for synthesis wrapper generation

kevindkim723:
Sometimes, we don't need/want to generate a wrapper for modules we want to synthesize, like modules in src/generic

Added a parameter to the makefile called WRAPPER=[0/1] that turns off/on wrapper generation respectively


ross144:
@kevindkim723 Can you add the SPDX header to  synthWrapper.tcl

PR #416, opened 2023-10-03
Update to Wally for test float fixes and other ancillary quality improvements

stineje:
Add update to test float test bench, Tex additions, Makefile for SoftFloat example and TeX for divsqrt that I recreated.


ross144:
There are output files (.aux, .log, .pdf) from pdflatex committed to the repo.  I'd prefer if these were not included.

PR #480, opened 2023-11-15
wrapper insertion automatically for Wally vs. individual PPA analysis

stineje:
Inserted wrapper into Makefile so can insert wrapper automatically which is needed for wrapperGen.py for inserting the configuration.

### DIFF
--- a/synthDC/Makefile
+++ b/synthDC/Makefile
@@ -1,28 +1,7 @@
-#####################
-# Makefile
-#
-# Written: ssanghai@hmc.edu, mmasserfrye@hmc.edu, james.stine@okstate.edu 15 November 2023
-#
-# Purpose: Makefile to be used for synthesis using DC
-#
-# A component of the Wally configurable RISC-V project.
-#
-# Copyright (C) 2021 Harvey Mudd College & Oklahoma State University
-#
-# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
-#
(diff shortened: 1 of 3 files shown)


davidharrishmc:
I'm concerned about this PR using WRAPPER to tell if something is a PPA run.  The version I wrote about two weeks ago checks whether there is a cvw_t parameter to decide if a wrapper is needed.  This PR sets wrapper to 1 by default, which will create a wrapper around a simple module with no parameters.  I think this will lead to several bugs that we won't discover until later.

I'd prefer that the PPA script just create a module such as add32 synthesize it over a range of target frequencies without messing with WRAPPER. I think this can be done from a PPA script entirely external to Makefile and synth.tcl so we could roll back to clean versions of those.  Every time Makefile or synth.tcl get touched, I have to revalidate the synthesis results and subtle things sometimes break that take a lot of time to sort out.

stineje:
This is what I tried to convey to you earlier, but, perhaps, I did not make it very clear.    The problem is due to the different methods for using synthesis for different approaches.  Wally will use the wrapper and add32 may not with the PPA script.  I am happy to fix this with the “same” parsing, but I will have to modify the python.  The other option which we discussed is just make it unique only to Wally and perform add32 outside of it.  However, I think this may be useful to have within the repository so I would like to get it integrated together.

I appreciate the explanation.  I will get this in by tomorrow morning/late tonight.

PR #488, opened 2023-11-19
FPGA Bootloader Preload From File

JacobPease:
- Linux Makefile: Fixed find utility crashing for real this time.
- Slight modification to testbench.sv
- Replaced vivado-risc-v addins directory with new SDC repo.
- Deleted vivado-risc-v directory and added ahbsdc.
- ahbsdc submodule actually added this time.
- Updated ROM to preload bootloader from file and infer a block ram when building for FPGA.


davidharrishmc:
The build root sims don't pass regression now because the boot.mem is missing.  Can we check a good version of this file into the repo, or is it critical to generate it dynamically?  If we need to generate, can you give me detailed directions (all the way back to pulling in the right version of Linux).

davidharrishmc:
Ah, the file was there. regression didn't like the path.  I've fixed it by changing the path in rom1p1r to be based on $WALLY.

  initial begin
    if (PRELOAD_ENABLED) begin
      $readmemh("$WALLY/fpga/src/boot.mem", ROM, 0);
    end
  end

PR #503, opened 2023-11-22
Fix testplan link

HepoH3:
Link to a testplan in the README is broken now. This PR fixes it.


davidharrishmc:
Thank you for the fix!

PR #553, opened 2024-01-06
Fixed testbench for coremark. Possibly broke verilator.

ross144:
- Fixed bug # 547, but there are other bugs which follow.
- added sig_end_canary to coremark.
- Fixes coremark. Maybe works with verilator.


davidharrishmc:
This fixed the CoreMark sim.  
With the signed int type, CoreMark Performance is up to 2.73 CM/MHz!  MINSTRET is down to 3135474.
Verilator still gets as far as it did before.  Not broken by this commit.
Whoo!

Issue #541, opened 2023-12-25
testbench-fp does not check CVTINT or CMP 


davidharrishmc:
At line 952, testbench-fp excludes checking CVTINT or CMP.

assign CheckNow = ((DivDone | ~divsqrtop) | (TEST == "add" | TEST == "fma" | TEST == "sub")) & (UnitVal !== `CVTINTUNIT) & (UnitVal !== `CMPUNIT);

CVTINT code later checks the flags but not the result.  There is no CMP check at all.

As a consequence, it is possible to change a bit in the expected answer (e.g. tests/fp/vectors/ui32_to_f128_rne.tv or f128_eq_rne.tv) and have the mismatch go undetected.

davidharrishmc:
I changed fp/vectors/ui32_to_f128_rne.tv to add ones near the right of lines 4 and 5, and sim-testfloat-batch cvtint still passed.

1FEFFFEF_401BFEFFFEF000000000000000000000_00
00000000_00000000000000000000000000000000_00
0000448E_400D1238000000000000000000000000_00
00000001_3FFF00000000000000000000000000**1**0_00
007FFFCD_4015FFFF340000000000000000000000_0**1**

stineje:
I did a full run on this and thought was testing everything.  Apologies - checking this now.

PR #562, opened 2024-01-12
Fixes for Issue #541

stineje:
Fix to compare both FlagMatch and ResMatch for conversions.  After fix, an additional bug was found on 754 invalids for conversion from fp to int.  All tests pass when run.


davidharrishmc:
According to Table 11.4 of the RISC-V spec, it looks like an out-of-range negative floating-point number converted to a uint32 should produce 0, not 2^31-1.

PR #568, opened 2024-01-13
replace $stop with $finish to avoid "Aborting... core dump"

solomatnikov:
replace $stop with $finish to avoid "Aborting... core dump"

### DIFF
--- a/testbench/testbench.sv
+++ b/testbench/testbench.sv
@@ -285,7 +285,7 @@ module testbench;
     if (TEST == "coremark")
       if (dut.core.priv.priv.EcallFaultM) begin
         $display("Benchmark: coremark is done.");
-        $stop;
+        $finish;
       end
     if(Validate) begin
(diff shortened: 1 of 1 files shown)


davidharrishmc:
Unfortunately, changing $stop to $finish terminates the Questa GUI simulations, which is a problem when debugging if I recall correctly from some time back.  Please just ignore the aborting error.

solomatnikov:
I don't know what issues Questa has but for robust verification and robust verification infrastructure things like `Aborted (core dumped) ./obj_dir/Vtestbench` should not be ignored.

`ifdef VERILATOR` can be added for verilator.

davidharrishmc:
$stop is the standard way to enter interactive debug mode.  See pages 592 and 1103 of IEEE_Std100-2017.  

I suggest filing a Verilator issue that it is nonstandard to abort and core dump on $stop.

Issue #585, opened 2024-01-20
Update the README.md for Tool chain installation


Sanket-0510:
This issue is for improvement in the tool chain installation guide for new comers. `sudo $WALLY/bin/wally-tool-chain-install.sh`  triggers all the installation command at  the same time this can cause system to run out of memory eventually crashing down, atleast my system crashed and stopped all the running processes. We can break down the [https://github.com/openhwgroup/cvw/blob/main/bin/wally-tool-chain-install.sh](url)  file in steps for the installation guide.

davidharrishmc:
See PR 586

PR #586, opened 2024-01-20
update the tool-chain installation guide

Sanket-0510:
This PR resolves - #585.
I will breakdown the `Walley-tool-chain-install.sh` into steps for installing different tool-chain so that sudden execution of 
`Walley-tool-chain-install.sh` won't cause any system failure. @ross144 Shall I go ahead with this improvement?


Sanket-0510:
@davidharrishmc I have attempted to update Readme.md for step by step tool-chain installation. I felt the need cause `walley-tool-chain-install.sh` file contains too much instructions making it difficult to read it and do one by one installation. Better we include it in Readme only.

davidharrishmc:
Hi, Sanket.  Thank you for looking at this.
I agree that invoking the entire wally-tool-chain-install.sh at once on an untested machine is unlikely to work well.  However, we've concluded it is better to cut and paste lines from the script than to cut and paste from the README. The script can be executable once a machine is robust (such as installing in a docker container in a way that's been known to work before), while pasting from the README is inherently manual.

PR #607, opened 2024-01-31
Adding nightly regression into the bin directory

Thomas-J-Kidd:


davidharrishmc (changes requested):
.gitignore, wally-tool-chain-install.sh, and Makefile changes should not be pushed into the repo.  Please remove these files from the PR.

Thomas-J-Kidd:
I believe I was able to remove them.

ross144:
Looks like Thomas applied your requested changes.  I think let's squash the pull request into a single commit to not have the unnecessary changes in the pubic repo.

PR #622, opened 2024-02-07
Finally have a regression-wally which has a single testbench.sv, a simple wally-batch.do, and also runs the branch predictor simulations correctly!

ross144:
- Updated to simplify configOptions.
- Updated deriv list to correct TWOBIT predictor configs.
- Finally have buildroot running in the main testbench!
- Finally have regression-wally doing everything!


ross144:
Major cleanup in wally-batch.do.
Still needs work to merge the imperas do files into wally-batch.do.
Also merges the imperas linux testbench into testbench.sv

PR #639, opened 2024-02-21
added softfloat support to regression-wally

kevindkim723:
running regression-wally with -softfloat flag runs sim-testfloat-batch for a select number of configurations


davidharrishmc:
@ross144 can you review that this plays nicely with your recent sim improvements?

ross144:
I think this is fine given the floating point testbench is a unit level test bench for the fpu rather than a top level Wally testbench.

PR #642, opened 2024-02-23
Improved alignment logic

ross144:
Also fixes nightly lint errors.
Still need to add support for extra wide floats on busn without cache.
I recall David said there are tests to check for quad on rv32gc and rv64gc but I couldn't find them in
tests.vh.
- Optimized the align logic for loads.
- Partially working optimized subwordwrite for misaligned.
- Closer to getting subword write misaligned working.
- Non-ideal fix.  Added new output from pma which indicates if the write shift should occur. The more ideal solution would be to have the pma indicate if the shift should occur and the maximum amount..
- Swapped to the more compact subwordreadmisaligned.sv.
- Beginning subword cleanup.
- Updated subword misaligned.
- Cleanup.
- Siginficant cleanup of subwordwritemisaligned.
- Simplifications of subword code.
- Removed duplicate endianswap.
- Cleanup.
- Added sdc to pma allow shift.
- Fixed lint errors for alignment.


davidharrishmc:
There are no quad tests yet, but rv32gc should stress things because it has doubles.  I'll open an issue to write wally64q to do some basic quad tests.

davidharrishmc:
AllowShift seems like a confusing name for the PMA to produce.  Is there something better?  Maybe SupportsMisalignedAccesses?

ross144:
SupportsMisalignedAccesses is not quite right either.  The control signal either allows subwordwrite to shift the WriteData or not.  For cached access, uncached memory access, clint, and sdc we need to position byte, half, and word into the correct byte offset.  However for the other peripherals we want to the keep the WriteData in the lsbs. Misaligned addresses are still not allowed for clint and sdc even though we need to assert AllowShift.

Perhaps AlignWriteDatatoOffsetM is a better name?

davidharrishmc:
I see (at least partly) how this has become such a mess.
The PMA checker ought to put out memory attributes that relate to the architecture, such as cachable, or atomic allowed, or word sizes allowed.  AlignWriteDatatoOffset isn't a very nice memory attribute.

I think the basic problem is that most peripherals can't accept misaligned or subword accesses.  Replicating the logic to support byte write enables on all the different registers could be painful, so we don't bother.

I think there's also a logic bug in the CLINT.  A write to any portion of the register starting at 0000 will change sip, even if it not encompassing the lsb.
        if (entry == 16'h0000) MSIP <= PWDATA[0];

What if we just prohibit misaligned accesses to all peripherals?  What if we do this by having another MisalignedSupported physical memory attribute that is only true for RAMs and ROMs? And can the aligner use such an attribute to do control what you presently need AllowShift for?

davidharrishmc:
Or maybe there is more to this?  I think it's worth getting to the bottom of it, rather than doing something hacky.

davidharrishmc:
Privileged spec 3.6.3.3 mentions an Alignment PMA that is not yet in our adrdecs.  Seems like this could be the time to add it.  If the Alignment PMA is 1, then misaligned accesses could cause an access fault exception instead of misaligned access.  This would be appropriate for the peripherals.

ross144:
The problem wit the CLINT and SDC aren't really an issue with misaligned access but rather aligned access smaller than the native width such as sb when XLEN=64.  These peripherals expect the byte to be shifted into the correct position within the word or double word.  The other peripherals do not expect this behavior because they only allow one sized access so the WriteData is required to be the lowest byte(s).  Our previous subwordwrite module duplicated the valid byte, half, or word but the new version shifts the valid data.  The Current AllowShiftM  control signal either enables or disables this shifting based the choice of peripheral. 

I'm missing how replacing this with a MisalignedSupported will work unless we allow the CLINT to support Misaligned access. But that doesn't make sense at higher level since the CLINT is inherently aligned.

Another completing issue is I re-synthesized the FPGA with these changes the results are disappointing.  The critical path is actually 1ns slower than the old misaligned despite having fewer levels of muxing logic.  The new subwordread costs 1 extra 2:1 mux but removes the 8:1 mux align shifter. The new subwordwrite costs a 4:1 mux but removes the 8:1 mux align shifter.   In total we should be an 8:1 mux shorter.  I suspect the reason for increased delay is the increased fanout.  The wire propagation time increased significantly.  I don't know how this will translate to 28nm.

Finally the new logic does pass regression and fpga linux boot.

davidharrishmc:
Should CLINT and SDC only allow preferred access size too? 

There's got to be a simple solution here.

PR #645, opened 2024-02-29
Add bsg_dmc memory controller and supporting infrastructure

infinitymdm:
## Summary of Changes
- Add basejump_stl as a submodule under the soc directory

- Add an asynchronous fifo to support crossing clock domains
- Add ahbxuiconverter.sv, which converts the Xilinx User Interface used by bsg_dmc to something AHB-compatible (theoretically, anyways)
- Add bsg_dmc_ahb.sv, which presents the memory controller with an AHB interface

## To Do Before Merging
- [x] Run regression tests
- [x] Simulate with the LPDDR model included with bsg_dmc


ross144:
Do we need an entire copy of basejump_stl or can we reduce it down to just the memory controller?

stineje:
Prob okay with just memory contr - i orig made basejump as a submodule but this is probably excessive.  Feel free to delete.

Sent from my iPhone

On Feb 29, 2024, at 8:52 PM, Rose Thompson ***@***.***> wrote:

﻿

Do we need an entire copy of basejump_stl or can we reduce it down to just the memory controller?

infinitymdm:
As I mentioned in the PR body, I have the basejump_stl submodule configured to sparse-checkout just the directories relevant to the memory controller. If you checkout my fork you should see that there as well.

infinitymdm:
Ok, it looks like it isn't just github not showing the sparse-checkout; the issue is that the sparse-checkout configuration lives in `.git/config`, which doesn't get committed, and git doesn't allow you to override the `submodule.$name.update` command for security reasons. AFAICT there is no way to _automatically_ sparse-checkout just the files we need. The options I'm aware of are:

- Keep basejump_stl as a submodule, and add `shallow = true` to gitmodules
    - **Pros**:
        - Already working
        - Makes IP ownership clear
        - Allows for simple configuration management using the typical `git submodule` flow
        - Roughly identical storage requirement to a proper sparse-checkout (we store extra .sv files, but drop most of the git history with the `shallow` flag)
        - Users can still do the sparse-checkout if they want by executing `git sparse-checkout set /bsg_clk_gen /bsg_dmc /bsg_misc/bsg_defines.sv /bsg_tag /testing/bsg_dmc/lpddr_verilog_model` from the basejump_stl directory (after submodule init)
    - **Cons**: 
        - Requires some knowledge of git features to update the submodule
        - We're stuck with the directory structure from BSG
- Remove basejump_stl as a submodule, and instead statically include the files we need
    - **Pros**: 
        - Simple; doesn't require any git-fu
        - Allows us to put these files wherever we prefer
    - **Cons**: 
        - We will need to be explicit about where the basejump files came from by adding copyright headers to all the files
        - Makes configuration management much more difficult in the case that bsg_dmc sees updates we want to include (now we have to go copy over all the files again, etc.)

I'm in favor of keeping the submodule, as I think it makes our collaboration with U of W and BSG clearest. Thoughts?

stineje:
I trust you - try what looks good.  Thx!!!!

Sent from my iPhone

On Mar 1, 2024, at 3:18 PM, Marcus Mellor ***@***.***> wrote:

﻿

Ok, it looks like it isn't just github not showing the sparse-checkout; the issue is that the sparse-checkout configuration lives in .git/config, which doesn't get committed, and git doesn't allow you to override the submodule.$name.update command for security reasons. AFAICT there is no way to automatically sparse-checkout just the files we need. The options I'm aware of are:

  *   Keep basejump_stl as a submodule, and add shallow = true to gitmodules
     *   Pros:
        *   Already working
        *   Makes IP ownership clear
        *   Allows for simple configuration management using the typical git submodule flow
        *   Roughly identical storage requirement to a proper sparse-checkout (we store extra .sv files, but drop most of the git history with the shallow flag)
        *   Users can still do the sparse-checkout if they want by executing git sparse-checkout set /bsg_clk_gen /bsg_dmc /bsg_misc/bsg_defines.sv /bsg_tag /testing/bsg_dmc/lpddr_verilog_model from the basejump_stl directory (after submodule init)
     *   Cons:
        *   Requires some knowledge of git features to update the submodule
        *   We're stuck with the directory structure from BSG
  *   Remove basejump_stl as a submodule, and instead statically include the files we need
     *   Pros:
        *   Simple; doesn't require any git-fu
        *   Allows us to put these files wherever we prefer
     *   Cons:
        *   We will need to be explicit about where the basejump files came from by adding copyright headers to all the files
        *   Makes configuration management much more difficult in the case that bsg_dmc sees updates we want to include (now we have to go copy over all the files again, etc.)

I'm in favor of keeping the submodule, as I think it makes our collaboration with U of W and BSG clearest. Thoughts?

ross144:
I think I'm ready to merge this into the SoC branch. But a few things I want to clean up.  This can occur after the merge.
1. I'd like to merge testbench-soc.sv with testbench.sv
2. I don't fully understand why we need a wally-soc.do. Can the SoC just be another configuration of wally? and then the regular wally.do and wally-batch.do scripts handle selecting the soc config.

infinitymdm:
> I don't fully understand why we need a wally-soc.do. Can the SoC just be another configuration of wally? and then the regular wally.do and wally-batch.do scripts handle selecting the soc config.

The main reason for an additional dofile was to include all the basejump files that aren't used in the usual wally configs. There are a bunch of svh files in different dirs, so there are a bunch of `+incdir+path/to/dir` bits that aren't in the normal dofile. This line in particular is the main change vs the normal wally-batch.do: https://github.com/infinitymdm/cvw/blob/ec48e11a547b632c73899812c3d1b4751fc61152/sim/wally-soc.do#L71

One solution could be moving files relevant to the soc effort into the src folder, so that they get picked up by the normal wally[-batch].do glob matching. We'll still have to tweak things to account for the svh files that basejump has all over the place, but it should allow us to proceed without a separate dofile for the SoC.

PR #649, opened 2024-03-06
Further simplified subwordread muxing

davidharrishmc:


ross144:
Isn't changing '0 to 0 going to cause a width mismatch? 0 is always 32-bits wide.

davidharrishmc:
It causes a width issue when passed as a parameter, but does the right thing when assigned to a variable.

Chapter 4 recommends using 0 instead of ‘0 and I figured the code should match the text.  0 also seems more readable.  It’s inconsistent with ‘1 or ‘x, but those are less common cases.

PR #659, opened 2024-03-08
Use correct version of riscv-gnu-toolchain and Fix Typo with ECA email

Karl-Han:
## Summary of Work

- As mentioned in #296, the master branch version is causing the problem of illegal operation as the change of binutils's version. I make modifications to change the `bin/wally-tool-chain-install.sh`.
- Found one typo in `Makefile`


ross144:
I thought it was the master branch of the gcc tool chain, not sail-riscv which caused issues with the la instruction.

Karl-Han:
As shown in the [submodule difference](https://github.com/riscv-collab/riscv-gnu-toolchain/compare/2023.12.20...2024.02.02) from the working version and the problematic one, I don't think gdb and binutils are causing the problem of assembly.

ross144:
I thought it was binutils which was causing the issue. Version 2.42 has the la bug.  tag 2023.12.20 of the gnu toolchain uses binutils version 2.41 but the later versions use 2.42.

davidharrishmc:
Why does this PR specify a certain version of Sail?

PR #663, opened 2024-03-11
Crypto Hardware Formatting and Simplification

davidharrishmc:
Lots of work to format the crypto extension units, and some minor optimization.
Added a few missing test cases.


davidharrishmc:
The changes now share logic between AES middle/final rounds and between encrypt and decrypt and with RV64 key schedule instructions, reducing AES area by more than 50%.

There is still a bit of optimization to be done.  One sbox should be shared between aes64e encryption and aes64ks1i key scheduling.

Also much of the sha logic can be shared between instructions, such as the XORs and sign extension.

PR #671, opened 2024-03-13
Increase number of jobs in riscof to speedup building.

Karl-Han:


ross144:
Is there a way we can make this platform dependent, similar to how make does --jobs?

Karl-Han:
I don't think so as I check the help from RISCOF:
```shell
$ riscof run --help
    INFO | ****** RISCOF: RISC-V Architectural Test Framework 1.25.3 *******
    INFO | using riscv_isac version : 0.18.0
    INFO | using riscv_config version : 3.14.3
Usage: riscof run [OPTIONS]

  Run the tests on DUT and reference and compare signatures

Options:
  --suite PATH     Path to the custom Suite Directory.  [required]
  --env PATH       Path to the env directory for the suite.  [required]
  --config PATH    The Path to the config file. [Default=./config.ini]
  --work-dir PATH  Path to the work directory.
  --no-browser     Do not open the browser for showing the test report.
  --dbfile PATH    The Path to the database file. NOTE: This argument is
                   mutually exclusive with  arguments: [testfile].
  --testfile PATH  The Path to the testlist file. NOTE: This argument is
                   mutually exclusive with  arguments: [dbfile].
  --no-ref-run     Do not run tests on Reference
  --no-dut-run     Do not run tests on DUT
  --no-clean       Do not clean work directory(if exists).
  --help           Show this message and exit.
```

And I think it should be done with `config.ini` since it is the main input for RISCOF. 

Maybe I will put all the configuration in one place, for example, `.env` with python to configure the all the tools, or global environment.

davidharrishmc:
Four sounds like a reasonable number of tasks given that most servers will have 8-16 cores now.

davidharrishmc:
38 minutes on chips before this PR.
12 minutes after.
Big improvement!

PR #672, opened 2024-03-13
Eliminate posedge clk

Karl-Han:
- $finish is required for Verilator while causing problem in QuestaSIM
    - use macro to conditional compile the $finish and $stop for different simulator
- Testbench is running much slower in verilator
    - My intuition for the modification is that I am trying to eliminate unnecessary operations, and I found that the loading and validating the test is included inside a always block.
    - as it turns out, this solution eliminates the problem and it runs faster on verilator than QuestaSIM


davidharrishmc:
Replaced by #704.

PR #673, opened 2024-03-14
Adding argparser to nightly_build.py in addition to minor updates

Thomas-J-Kidd:
nightly_build.py got a makeover through adding logger and argparser. I continued to add try and except statements to help prevent crashes and no error logging. 

argparser now allows you to specify path, repository, which tests to make, and to send an email or not. 

I edited the README.md to show how to add a cronjob

I updated the wrapper script to match the new argparser functionalities. 

Please squash my commit because some of them are old and reverted.


davidharrishmc:
What do you mean by squash your commit?  Do you want this PR committed?

Thomas-J-Kidd:
I though there was a feature where my many commits can be combined into just 1? I only said this because some of my commits are merges or reverts which dont really add value to the commit history. I am not sure how important that is, thats why I mentioned it. 

If it is not a big deal, then its okay. I would love this PR to be committed :) !

ross144:
Squash means converting all of the commits in the pull request into a single commit.  I'm generally not a fan of squashing because it hides the true history of how the code was developed.  However if there are truly undesired intermediate changes which are not part of the final commit then squashing can be useful.  For example if a large fill was accidentally added and then removed it would still be part of the history and take up space in the git repo.  Since squashing combines all the commits into a single commit its it behaves as if the file was never added.

For this pull request I prefer we just merge.

PR #677, opened 2024-03-16
VCS 

Divya2030:


davidharrishmc:
Divya, this PR has more than 100 commits to 94 files.  I don't think you are trying to change all these files.  It looks like they are an artifact of a merge that you have done.

Could you cancel this PR and reissue it with just the changes you are intending to make?  

You may need to do a "git pull upstream main" to update your clone of your fork, and then apply the changes you intend, and then commit those changes.  Let me know if you have any problems.

davidharrishmc:
Also, it's good to make the PR into the main branch of CVW instead of a vis branch, as long as you've tested you won't break anything. You may have to close this PR and reissue to the main branch.

ross144:
This is my fault i created the vcs branch as I didn't want to merge in that temporary file with the empty module.  We should synth this branch. I'll make a pull request and then the diff will look more reasonable.

ross144:
Ok I made the fast forward pull request. This should reduce the number of diffs in this pull request.  If we are happy with your changes.  Lets go ahead and merge into the main branch.

davidharrishmc:
Now there's a conflict, and I think we need to change the target to the main branch.

ross144:
The only thing I see missing the copy right and authorship on the two script files. Can you add these? You can base this off of any of the system verilog file?

ross144:
And yes I now agree with David we should just change the target to the main branch.

PR #718, opened 2024-04-08
Putting Toolchains and Regression inside Docker with Scripts

Karl-Han:
There are two parts in this PR:

- Dockerfiles/Container/Commandline for
    - Open-Source Toolchains including buildroot
    - Automatic Regressions within Docker
- Easy-to-use scripts
    - main start-up script: `start.sh` to start the command line within container
    - assistive scripts: not intended to be used directly but for reference
        - `get_images.sh`: get necessary Docker images by fetching (default)
        - `run_regression.sh`: used in automatic regression

See [Dockerfiles README](https://github.com/Karl-Han/cvw/tree/docker/docs/docker#readme) for detailed documentations.

Please try it out and give me some feedbacks on better documentations and easier start-up script.


ross144:
Why does docker need a copy of the buildroot configs and many of the linux script files?

Karl-Han:
It is impossible to make docker to follow the symbol link and copy the files.
> https://stackoverflow.com/questions/31881904/docker-follow-symlink-outside-context

Therefore, I wrote a short python script `get_buildroot_testvector.py` to copy those folders from relative path if you want to build it. And `get_images.sh` is the script that will invoke it if the user explicitly specifies building `wallysoc/toolchains_wally`.

PR #722, opened 2024-04-09
Fixed path in run-imperas-linux.sh to match new sim structure

slmnemo:


davidharrishmc:
For now, just move run-imperas-linux into the questa directory.  It's questa-specific, so it should go in that simulator's directory.

Soon @ross144 is planning to unify the Linux testbench and this will go away.

PR #737, opened 2024-04-16
Fixed code coverage for regression-wally

ross144:
- Fixed insert_debug_comment.sh to work with the older version of bash.
- Patch makefile in embench to use wsim.
- Fixed regression-wally so it actually produces covereage reports.
- Fixed makefile and regression-wally so that code coverage now works.

### DIFF
--- a/bin/regression-wally
+++ b/bin/regression-wally
@@ -208,7 +208,7 @@ def addTests(tests, sim):
             gs = test[3]
         else:
             gs = "All tests ran without failures"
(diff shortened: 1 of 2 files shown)


davidharrishmc:
This is good to have.

Do you think questaCoverage belongs in another Makefile within the sim/questa directory?

davidharrishmc:
Can you make an issue or track somehow.  Merged.

PR #738, opened 2024-04-17
Added full Linux boot to regression-wally

slmnemo:
Added logfile and overwrite variables to TestCase tuple to decouple log output from files regression searches. Searchfile added to testcase list format, creating tests format 

`[config, name, args, grepstr, log location]` 

We may prefer to add the overwrite to this tests format with the log location such that

`[config, name, args, grepstr, [log location, overwrite]]`

Instruction Limit is currently at 600M. Not sure if this is adequate for a full Linux boot.


davidharrishmc:
What motivated changing INSTR_LIMIT to 4000000?  Does that make buildroot the slowest test suite during normal regression?  We don't want to slow down regression if we can help it.

Some of the regression testing had been hacky and difficult to understand.  We tried to do a bunch of cleanup.  Think carefully as you are extending it so we don't get hacky again.

tests_buildroot is confusing in that it's not distinguished from the regular buildroot sim.  Would it make sense to rename to tests_buildrootboot?

I think there are two logfiles involved in simulations with peripherals.  One reports all of the stuff that would normally go to the screen during sim, and the other is just UART output.  Your new code isn't clear to me whether it is related to UART, both, or neither.

slmnemo:
Ok, so I thought about it and I think this implementation should be much better. I've separated the logfile searched from the sim_log when a specific logfile is input. Otherwise the logfile just receives sim_log in addTests(). 

I also moved the log piping to the cmd defined in each testcase from where it was in run_test_case(). This adequately separates the logfile from the simulation log allowing us to search it.

PR #739, opened 2024-04-17
Add extra path to search for deriv/buildroot

Karl-Han:


davidharrishmc:
I like where this is going.  

Should the config files also be in the DEPENDENCIES?  Can we refactor DEPENDENCIES so that the same files aren't typed twice there and in INCLUDE_PATH or SEARCH_PATH such that we might change one and forget to change the other?

Karl-Han:
- DEPENDENCIES are specific configuration files that any modification will incur recompilation.
- INCLUDE_PATH are paths that the compiler should search for files.

These are different things. Even though there are some intersections between these two and it is possible to do so, it complicates things. Instead, writing comments or docs for the meaning is a better choice, and those who want to make modifications should be able to achieve their goal by reading comments and docs.

Issue #723, opened 2024-04-10
Accessing Environment Variable in rom1p1r.sv


europe2024:
Below is the platform.

     Verilator 5.025 devel rev v5.024-5-g1012c054e
     variant of Debian (version 12) on ARM

The following line

      $readmemh("$WALLY/fpga/src/boot.mem", ROM, 0);

in

     $WALLY/src/generic/mem/rom1p1r.sv

uses environment-variable $WALLY, but Verilator does not automatically generate code to get its value.

Below is one way to access $WALLY.

     rom1p1r.sv
     ----------
     import "DPI-C" function string getenvval(input string env_name);

     $readmemh({getenvval("WALLY"), "/fpga/src/boot.mem"}, ROM, 0);

     clangwrapper.c
     --------------
     #include <stdio.h>
     #include <stdlib.h>

     #include "Vtestbench__Dpi.h"

     const char *getenvval(const char *pszName) {
       return ((const char *) getenv(pszName));
     }

Karl-Han:
Thanks for your advice. We are currently working on getting Verilator working in the team, and your insight to have C++ execution is great.

I will notify you when the feature is implemented and close the issue afterwards.

Karl-Han:
I tried to replicate your idea with passing RISCV in `testbench/testbench.sv`, but I cannot resolve the problem of undefined symbol. 

testbench/testbench.sv:
```systemverilog
import "DPI-C" function string getenvval(input string env_name);
string       RISCV_DIR = getenvval("RISCV"); // "/opt/riscv";
```

Log:
```shell
mold: error: undefined symbol: Vtestbench__ALL.a(Vtestbench___024unit__DepSet_hc0e221bc__0.o): getenvval
collect2: error: ld returned 1 exit status
make[1]: *** [Vtestbench.mk:66: Vtestbench_rv64gc] Error 1
```

Here are the questions I have:

- Can you share your command to compile the executable with Verilator?
- You are wrapping getenv with getenvval to avoid confliction in declaration of getenv in C and C++, right?

## Verilator Compilation Command

```shell
time verilator \
--Mdir obj_dir_non_profiling -o Vtestbench_rv64gc \
-cc --binary \
 -DVERILATOR=1 --no-trace-top --stats \
--timescale "1ns/1ns" --timing --top-module testbench  --relative-includes \
"-I/home/khan/cvw/config/shared" "-I/home/khan/cvw/config/rv64gc" \
wrapper.c \
[...]

PR #745, opened 2024-04-18
Add getenvval for rom. Related to issue #723.

Karl-Han:


davidharrishmc:
I haven't been following this closely.  Does this fix work for Verilator, Questa, and VCS? Has it been tested?

PR #748, opened 2024-04-22
FPGA fixex required after regress updates and added compatiblity for cygwin.

ross144:
- Fixed fpga to work with the updated regression changes.
- Maded insert_debug_comment.sh compatible with cygwin.


ross144:
@davidharrishmc Can you merge?  I need these changes for my class.

PR #777, opened 2024-05-04
directed coverage tests, vcs partially running

davidharrishmc:


ross144:
@davidharrishmc There are merge conflicts.  I'm not sure the correct solution for covergen.py and covergen_footer.S.

PR #797, opened 2024-05-14
Update MISA B bit and riscvassertions for B and C bits based on spec

jordancarlin:
- Based on the spec, B should be set if Zba+Zbb+Zbs are enabled. This was added to relevant configs and riscvassertions was updated to ensure B is set if these extensions are set (and vice versa)
- Based on the spec, C should be set if Zca + Zcf (if RV32 and F supported) + Zcd (if D supported) are enabled. This was added to riscvassertions to ensure C is set if these extensions are set (and vice versa).
- C_SUPPORTED and COMPRESSED_SUPPORTED were replaced with the relevant sub extension for conditional generates

This keeps the current config setup and just adds the B MISA bit when needed and adds some new assertions to ensure compliance with the spec when setting config files. These new assertions got kind of complicated. There are also now a lot of ways to create configs that are not supported and when enabling certain extensions they need to be set in multiple places in the config. An alternative approach that avoids these problems (though introduces more significant changes) is introduced in #798.


davidharrishmc:
If we go with this method, derivlist.txt will need updates on MISA entries to match.

jordancarlin:
> If we go with this method, derivlist.txt will need updates on MISA entries to match.

Yes. That's another reason I prefer the other version. If we change if an extension is supported in a main config the derived configs will get that change automatically since MISA is not being redefined.

jordancarlin:
Closing in favor of #798.

PR #798, opened 2024-05-14
Update config to derive MISA from macros and update MISA bits based on the spec

jordancarlin:
The config files and config-shared were modified so that in the config files individual extensions are enabled and config-shared derives the MISA field from the supported extensions. All of the main configurations + all the derived configurations were updated. Nightly regression is passing with these changes.

The following extensions are set automatically if their relevant sub-extensions are set:
- A_SUPPORTED from A sub extensions
- B_SUPPORTED from B sub extensions:  Based on the spec, B should be set if Zba+Zbb+Zbs are enabled
- C_SUPPORTED from C sub extensions:  Based on the spec, C should be set if Zca + Zcf (if RV32 and F supported) + Zcd (if D supported) are enabled.
- ZKN_SUPPORTED from Zk* sub extensions

Additionally, the following changes were made to other extensions/conditional generates/assertions:
- C_SUPPORTED and COMPRESSED_SUPPORTED were replaced with the relevant sub-extension
- M implies Zmmul, so is required if M is set

This approach to the config files eliminates a lot of the potential to set certain parts of extensions without setting the MISA bit, which allows many assertions to be removed. This reduces the chances of getting failures when creating new configs. Additionally, especially for the derived configs, it is easier to see what is changing for each configuration, as the entire MISA field doesn't need to be set from scratch again. This also makes further changes much easier as new extensions are ratified that may impact MISA bits.

This is an alternative to #797.


davidharrishmc:
Where in the spec does it say

C should be set if Zca + Zcf (if RV32 and F supported) + Zcd (if D supported) are enabled.

jordancarlin:
screenshot_2024-05-12_at_10 51 03___am_720

PR #816, opened 2024-05-28
Merges support for functional coverage into wally.do and testbench.sv

ross144:
This is bit hacky and will need to be cleaned up with future commits.  However, I have the base features finally working.  ImperasDV and riscvISACOV both work with wally.do, wsim, and testbench.sv and don't require any outside scripts.

To run functional coverage add the flag --fcov to wsim and run with a single elf file.

I've confirmed coverage is generated in the GUI.


davidharrishmc:
Merged.  Please continue cleanup and remove testbench-imperas.sv if it is no longer needed.

PR #823, opened 2024-06-04
Debug Support seed

stineje:
Initial support for debug specification.  Has minimal configuration.  Will post documentation shortly but seems to work.  Lots of changes here so let us know if we messed anything up.  Apologies for the delay too.  I know it will take time to clean up and get in so please let us know how we can make better.


bin/hw_debug_interface.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+def write_data(register, data):
+    """Writes data of width XLEN to specified register"""
+    # Translate register alias to DM regno
+    regno = int(register_translations[register], 16)
+    # Write data to 32 bit message registers
+    data = int(data, 16)
+    write_dmi("0x4", hex(data & 0xffffffff))
+    if XLEN == 64:
```
jordancarlin: Shouldn't this be `XLEN >= 64`?

bin/hw_debug_interface.py
```diff
@@ -0,0 +1,397 @@
...
+# TODO: if JTAG clk is fast enough, need to check for busy between absract commands
+
+def dump_GPR():
+    gpr = {}
+    for i in range(1,32):
+        addr = f"X{i}"
+        gpr[addr] = read_data(addr)
+        # DM will assert Abstract Command Err if GPR X16-X31 isn't implemented (CMDERR_EXCEPTION)
+        # This will clear that error and return early. 
+        if i == 16:
```
jordancarlin: Is there any way to get the value of `E_SUPPORTED` and base the for loop range off of that to eliminate the need for this error handling? It might not be possible in the python script, but if it is it would simplify things.

config/shared/config-shared.vh
```diff
@@ -16,6 +16,9 @@ localparam SVMODE_BITS = (XLEN==32 ? 32'd1 : 32'd4);
 localparam ASID_BASE = (XLEN==32 ? 32'd22 : 32'd44);
 localparam ASID_BITS = (XLEN==32 ? 32'd9 : 32'd16);
 
+// Debug Module implemented                                                                                                          
+localparam logic DEBUG_SUPPORTED = 1'b0;
```
jordancarlin: This should probably be in the individual config files. Everything in the `config-shared.vh` file is either a constant or is algorithmically determined. 

Debug support can technically be turned on for different configurations (and we may want to do so using derived configs), which requires it to be in the individual config files. I suggest adding it after all of the ISA options but before all the microarchitecture features.

jordancarlin:
The changes made to the Arty 7 FPGA implementation probably also need to be made to the vcu FPGAs.

bin/hw_debug_interface.py
Matthew-Otto: Yes it should

bin/hw_debug_interface.py
Matthew-Otto: If the value of E_SUPPORTED is contained in the MISA (it seems like it might be, but I need to refer to spec), we could infer from that.

Matthew-Otto:
The python scripts are hacked together and just used as a quick way to verify the debug module functions in hardware. They may not belong in the parent repo. I'm investigating more elegant solutions (GDB) for future versions. The HDL features are the important aspect of this PR.

infinitymdm (commented):
You've got 4 variables used before their declarations in the file src/debug/dm.sv:
- `CmdErr`: used on L117, declared on L152
- `Busy`: used on L117, declared on L151
- `InvalidRegNo`: used on L120, declared on L165
- `RegReadOnly`: used on L120, declared on L166

This causes errors in some of our tools (notably ModelSim) while other tools (like verilator) don't care.

PR #836, opened 2024-06-18
Code Cleanup: Lint Improvements

davidharrishmc:
Verilator lint -Wall now runs with a few warnings turned off.  This would have caught the blocking LRU issue.
Cleaned up many unused signals.


src/uncore/uartPC16550D.sv
rosethompson, line 1: I didn't review the fifo empty and entries signals carefullly. Are we sure this does not introduce a bug. I don't want to find this in the FPGA.

ross144 (commented):
I want to take a closer review of the uart changes before I merge.

PR #851, opened 2024-06-22
Reduce Bit widths for IDIV on FPU

kevindkim723:
IDIV datapath on fpu uses divb so in some configurations the bit widths are unnecessarily wide. Created new parameters and bit widths to deal with this.


davidharrishmc:
looks like INTRESBITS, INTFPDUR are only used in config-shared.vh and INTDIVBLEN is not used at all.  Can you delete INTDIVBLEN entirely and remove the other two from paramdefs/cvw?

calparam INTRESBITS     = XLEN + LOGR; // number of bits in a result: r integer + XLEN fractional
localparam INTFPDUR       = (INTRESBITS-1)/RK + 1 ;                 
localparam INTDIVb        = INTFPDUR*RK - LOGR;                     
localparam INTDIVBLEN     = $clog2(INTDIVb+1);

PR #855, opened 2024-06-27
FPU without privilege modes + derived config fixes

### LINKED ISSUE #840: FPU fails when supervisor mode is not supported
When `S_SUPPORTED` is set to 0 and `F_SUPPORTED` is set to 1, the archXXf tests fail and get stuck in a loop of illegal instructions (as seen below)

`Program fetched illegal instruction 0x00000000 from address 0x00000000.  Might be fault with no fault handler.`

Floating point should not be dependent on privilege modes being supported, just Zicsr. None of the current derived configurations test with s/u modes disabled (other than the ones based on simpler configs like rv32imc), which is likely why this has not been discovered.

jordancarlin:
- Fix derived configs with `D_SUPPORTED` after #853 stopped `ZCD_SUPPORTED` from being turned off
- Add derived configs for no privilege modes
- Fix #840 FPU failing when `S_SUPPORTED = 0` by adjusting logic for updating `STATUS_FS`

### DIFF
--- a/bin/regression-wally
+++ b/bin/regression-wally
@@ -123,18 +123,14 @@ derivconfigtests = [
         ["zknh_rv64gc", ["arch64i", "arch64zknh"]],
 
 # No privilege modes variants
-        ["noS_rv32gc", ["arch32i", "arch32f", "arch32d", "arch32priv", "arch32c", "arch32m", "arch32a_amo",
-                        "arch32zifencei", "arch32zicond", "arch32zba", "arch32zfh", "arch32zfh_fma", "arch32zfh_divsqrt",
-                        "arch32zfaf", "arch32zfad", "wally32a_lrsc", "arch32zcb", "arch32zbkx", "arch32zknd"]],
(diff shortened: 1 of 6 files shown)


davidharrishmc:
Good discoveries.

The suite of tests run with no privilege derived configs is pretty large and includes some of the time-consuming ones.  Would it make sense to trim down a fair amount?  For example, is arch32f sufficient to show floating-point works without privilege? 

 ["noS_rv32gc", ["arch32i", "arch32f", "arch32d", "arch32priv", "arch32c", "arch32m", "arch32a_amo",
                        "arch32zifencei", "arch32zicond", "arch32zba", "arch32zfh", "arch32zfh_fma", "arch32zfh_divsqrt",
                        "arch32zfaf", "arch32zfad", "wally32a_lrsc", "arch32zcb", "arch32zbkx", "arch32zknd"]],

jordancarlin:
I wanted to keep it fairly large so it covers most areas since the initial issue came in such an unexpected place, but I went ahead and trimmed down floating point to just include the standard f tests.

PR #856, opened 2024-06-27
Testbench cleanup

jordancarlin:
- Use `VCS` instead of `SIM_VCS` for vas related ifdefs in testbench as `VCS` is enabled by default in the simulator and does not need to be passed as an argument
- Base `$stop` vs `$finish` off of the `QUESTA` macro instead of `VCS` and `VERILATOR` since Questa seems to be the odd one out and this only requires one ifdef instead of two
- Correct usage of `S_SUPPORTED` in testbench_fp and replace with `F_SUPPORTED`


davidharrishmc:
Great cleanup!

PR #862, opened 2024-06-29
Remove Verilator hack

jordancarlin:
It looks like the issue with Verilator that necessitated the workaround in #704 has been resolved in the latest release of Verilator: https://github.com/verilator/verilator-announce/issues/70. This reverts it to cleanup the testbench.

Do not merge until we update Verilator on chips and any other servers that are running Verilator (this should be much easier once the new installation script is pushed in the next couple days).


davidharrishmc:
Confirming this is ready to pull?

jordancarlin:
Yes. Should be good to go.

PR #866, opened 2024-07-03
First version of iterelf running; removed directory support from wsim

davidharrishmc:
wsim now requires --elf <elffile> when running an elf file.  The suite argument is used to uniquely name the work and log files so that multiple elf files can be simulated concurrently with iterelf.

The testbench now automatically makes the memfile, label, and address files if they are out of date or don't exist.  This saves the hassle of doing it manually for a random .elf.

iterelf has a max of 8 concurrent processes because of a limited number of ImperasDV licenses.

Example:
iterelf tests/riscof/work/riscv-arch-test/rv64i_m/I/src --sim vcs


jordancarlin:
The README section on wsim should probably be updated to reflect the need for --elf if running an elf file and the removal of directory capabilities. Not sure if we also want to add iterelf to it.

PR #869, opened 2024-07-04
Installation and setup overhaul

jordancarlin:
This introduces a completely overhauled installation script and accompanying setup scripts. The major changes are listed below:
- Compatibility with Ubuntu 20.04, 22.04, and 24.04 along with RHEL/Rocky Linux/AlmaLinux 8 and 9
- New `wally-distro-check.sh` script introduced to check which distro and version are being used and set environment variables accordingly. This is called by the main installation script
- Package installation from package manager split into separate script (`wally-package-install.sh`) so it can optionally be run separately by a system administrator. The packages are also split up by which tool needs them for better maintainability.
- The main installation script can now be rerun to update all tools or complete an installation that was interrupted. It will only reinstall tools that have newer versions or are missing.
- The main installation script can now be run at a user-level without root/sudo privileges. If run as root, it will set `$RISCV` to `/opt/riscv` and install system packages. If run as a non-root user, it defaults to `~/riscv` and just checks that all required packages are installed (run `wally-package-install.sh` as root first)
- Everything is installed to `$RISCV` instead of also having components installed to `/usr/bin` and several other directories
- Improved status and error messages are printed throughout the installation process. Limited messages are printed to the console and logs are produced for the installation of each tool.
- Python is switched to use a virtual environment. This allows for a different version of Python from the system version and keeps the installed packages isolated from the system. It also removes the need for sudo when installing pip packages.
- Python packages are installed from a `requirements.txt` file which allows for specific minimum versions to be specified
- All Python script shebangs were updated to use the Python virtual environment
- .gitmodules file updates to reflect current submodules and eliminate warnings when cloning
- README updated with new installation flow and general cleanup
- `setup.sh` updated to automatically select `$RISCV` directory if possible and always source `site-setup.sh` from it
- `site-setup.sh` updated to activate python virtual environment when run. It is also now automatically downloaded to the `$RISCV` directory and additional lines are automatically added to it to activate newer versions of gcc on certain distros.
- All hardcoded references to /opt/riscv are removed
- The Linux makefile is updated to generate the testvectors when running `make all` and to remove the need for separate sudo and no-sudo versions. Sudo is automatically invoked if needed to write to the `$RISCV` folder.
- Linux readme updated
- Main installation script now installs Buildroot Linux and generates testvectors.
- `--clean` flag added to installation script for space-constrained systems

Currently, Verilator fails some simulations on Ubuntu 20.04 and Red Hat 8. verilator/verilator#5295


davidharrishmc:
This is a great piece of work.  Let's get it into the repo and let people try it.  We can continue to refine if need be.  Let me know when you want me to merge.

PR #902, opened 2024-08-09
Build zsbl in main Makefile

jordancarlin:
Remove old boot.mem and generate boot.mem and data.mem with Makefile


davidharrishmc:
@ross144 could you confirm that we want to make boot.mem on the fly and not have it in the repo?

PR #907, opened 2024-08-10
Update linker scripts to avoid hardcoded /opt/riscv

### LINKED ISSUE #905: Remove hardcoded /opt/riscv from linker scripts
The only remaining hardcoded references to `/opt/riscv` are in the linker.x files. They don't support environment variables in them so an alternative approach needs to be found for passing in the SEARCH_DIR.

jordancarlin:
Remove `SEARCH_DIR` from linker scripts and pass `$RISCV/riscv64-unknown-elf/lib` as a command line flag in the makefile instead. Fixes #905


davidharrishmc:
I like this.  @ross144 please review/test/accept if it doesn't break anything for you.

PR #908, opened 2024-08-11
Sim updates + rv64gcCacheSim.py fixed

### LINKED ISSUE #767: rv64gc_CacheSim.py
This script is no longer working since our updates to the simulation scripts.  We need to update the script to work with vsim and the updated directory structure.

jordancarlin:
### CacheSim Updates
- Update rv64gcCacheSim.py to work with wsim and allow any simulator to be selected (fixes #767)
- Update rv64gcCacheSim.py to automatically override the `I_CACHE_ADDR_LOGGER` and `I_CACHE_ADDR_LOGGER` parameters. This ended up leading to a lot of other changes to the simulation scripts, described below
- Update cacheSim scripts to return the number of mismatches

### wsim Updates
- Add a new `--param` argument to wsim that takes an input in the form `"param1=value1 param2=value2 ..."`. This is used to convert to the correct format for parameter overrides in each sim-specific script and is used at the compilation stage, unlike the `--args` plus args which are used at run time.
  - Switch to using this method for overriding the TEST parameter for testbench_fp

### Questa Updates
- Use tcl list expansion to avoid the need for temporary variables and many small variables that always go together in wally.do
- Refactor argument parsing to work with both `--args` and `--params`

### VCS Updates
- Add testbench_fp support to VCS (currently failing with mismatched signature)
- Update VCS script to allow for parameter overrides to be passed as arguments
  - VCS uses a different syntax for overriding parameters, so the value of `--params` is modified and written to a file that VCS reads

### Verilator Updates
- Update Verilator Makefile to allow for parameter overrides to be passed as arguments


davidharrishmc:
I like this too.  @ross144 can you confirm this doesn't break the FPGA or anything else you use?

ross144 (approved):

ross144:
I'm happy with these changes. I may have to refactor a few things after the merge to get fpga rvvi in simulation again, but that's is ok with me.

ross144:
@jordancarlin Can you resolve the merge conflict in wally.do? Then we can merge. My apologizes for taking so long to review this pull request.

davidharrishmc:
@jordancarlin the wally.do has gotten stale.  I think we want the new one, but can you check and resolve the merge conflict?

jordancarlin:
@davidharrishmc @ross144 Conflicts resolved and should be good to go

PR #916, opened 2024-08-15
FPGA Makefile refactoring

jordancarlin:
@ross144 If you're interested, this new version of the makefile allows you to run make for the different FPGA boards without having to comment out a variety of different lines. You can either just change the `all` target to point at the board you want or run `make [ArtyA7|vcu108|vcu118]` directly from the command line to compile for a specific board. I've kept the default board as the ArtyA7.

### DIFF
--- a/fpga/generator/Makefile
+++ b/fpga/generator/Makefile
@@ -1,43 +1,42 @@
 dst := IP
 
-# vcu118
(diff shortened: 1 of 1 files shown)


ross144:
This is wonderful.

PR #919, opened 2024-08-16
Set RISCOF jobs based on number of cores

jordancarlin:
Automatically scale number of RISCOF jobs based on system core count. On chips this results in a 2x speedup from 12 minutes down to 6 for riscof.


ross144:
This is wonderful. Great work!

PR #933, opened 2024-08-27
Committing the custom test spitest.

JacobPease:


tests/custom/spitest/Makefile
```diff
@@ -0,0 +1,112 @@
...
+
+TARGETDIR	:= bin
+TARGET		:= $(TARGETDIR)/spitest.elf
+ROOT		:= ..
+LIBRARY_DIRS	:= 
+LIBRARY_FILES	:=
+
+MARCH           :=-march=rv64imfdc
+MABI            :=-mabi=lp64d
+LINK_FLAGS      :=$(MARCH) $(MABI) -nostartfiles
```
jordancarlin, line 26: It's interesting that this is working without it, but since removing `SEARCH_DIR` from the linker scripts, we've been adding ` -L $(RISCV)/riscv64-unknown-elf/lib` to `LINK_FLAGS` so that it is still able to search for libraries in that directoy.

JacobPease: Ah, thanks for pointing that out. It's actually not working. It had been a while since I compiled it. Working on it now.

PR #937, opened 2024-08-29
Fixed questa and ImperasDV library issue

ross144:
Fixes the 32-bit vs 64-bit library issues between Questa and ImperasDV
Modifies the FPGA to support SPI debugging.  Note the FPGA is in flux and may require some specific tweak to build correctly.


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "cvw-arch-verif"]
 	path = addins/cvw-arch-verif
 	url = https://github.com/openhwgroup/cvw-arch-verif
+[submodule "addins/riscvISACOV"]
+	path = addins/riscvISACOV
+	url = git@github.com:riscv-verification/riscvISACOV.git
```
jordancarlin: Should this use the `https` url like all of the other submodules?

rosethompson: Yes it should.

rosethompson: Just fixed.

PR #941, opened 2024-08-30
Add Imperas lockstep verbose as wsim option

### LINKED ISSUE #871: Restore verbose mode for ImperasDV
We've lost the ability to run ImperasDV in verbose mode now that we transitioned to wsim.

It seems that we used to do it by passing OTHERFLAGS= *** to vsim in wally.do.

Some scripts such as setup.imperas.sh or sim-imperas set these as

OTHERFLAGS="+TRACE2LOG_ENABLE=1 VERBOSE=1"

I wasn't successful trying this on the questa wally.do file

Adding (or uncommenting the line with)
--trace
--tracemode
to imperas.ic gives us the tracing back, but it would be good to be a wsim option.

jordancarlin:
Adds a `--lockstepverbose` flag to wsim that runs lockstep in verbose mode. `--lockstepverbose` automatically implies `--lockstep`. It's a bit hacky and uses a second `imperas-verbose.ic` file that is added to the main `imperas.ic` file when `--lockstepverbose` is passed, but was the cleanest solution I could find.

Fixes #871


ross144:
I agree it's a bit awkward. What is the difference between locksteplog and lockstepverbose?

jordancarlin:
Not totally sure what locksteplog does (@davidharrishmc), but I believe it sets how many instructions should pass before it begins logging. lockstepverbose just sets the flags that we've been toggling on and off with comments to enable more in depth output from imperas.

PR #942, opened 2024-09-03
Improved automated support for FPGAs

ross144:
Automatically sets the following based on config and makefile settings
1. Sets the zsbl cpu clock frequency (SYSTEMCLOCK)
2. Sets the max SPI clock frequency
3. Inherits the location and size of main memory from the fpga config.
4. The FPGA clock (mmcm) generator scripts now set the clock speed based on SYSTEMCLOCK

Todo:
1. Still need to fix makefile dependencies so the FPGA script will regenerate the mmcm if the clock changes.  Not sure how this should be done.
2. Finally update the device tree.  I think we can return to a single device tree file and have this Makefile create the correct dts for this particular build.


fpga/generator/Makefile
```diff
@@ -57,15 +65,19 @@ PreProcessFiles:
...
 	./insert_debug_comment.sh
-	# modify config  *** RT: eventually setup for variably defined sized memory
-	#sed -i "s/EXT_MEM_RANGE.*/EXT_MEM_RANGE = 64'h0FFFFFFF;/g" ../src/CopiedFiles_do_not_add_to_repo/config/config.vh
 	# This line allows the Bootloader to be loaded in a Block RAM on the FPGA
 	sed -i "s/bit \[DATA_WIDTH-1:0\].*ROM.*/(\* rom_style=\"block\" \*) &/g" ../src/CopiedFiles_do_not_add_to_repo/generic/mem/rom1p1r.sv
 	sed -i 's/$$WALLY/\.\.\/\.\.\/\.\.\//g' ../src/CopiedFiles_do_not_add_to_repo/generic/mem/rom1p1r.sv
 	sed -i 's/$$WALLY/\.\.\/\.\.\/\.\.\//g' ../src/CopiedFiles_do_not_add_to_repo/generic/mem/ram1p1rwbe.sv
 
+# build the Zero stage boot loader (ZSBL)
+.PHONE: zsbl
```
jordancarlin: Should be `.PHONY`

fpga/generator/ddr4-vcu118.tcl
```diff
@@ -1,11 +1,14 @@
 
 set partNumber $::env(XILINX_PART)
 set boardName $::env(XILINX_BOARD)
+set SYSTEMCLOCK $::env(CLOCK)
```
jordancarlin: `set SYSTEMCLOCK $::env(SYSTEMCLOCK)`

ross144:
Thanks for catching theses.  They are fixed now.

PR #945, opened 2024-09-05
Coverage updates

davidharrishmc:


addins/cvw-arch-verif
jordancarlin, line 1: Possible it's just an issue with git, but it is claiming that you deleted the `cvw-arch-verif` submodule

rosethompson: We don't want to remove this submodule.  We are not sure why this happened.  Correct me if I'm wrong, but this would happen if the .gitmodule file removed the submodule.  I don't see that change in the PR.

jordancarlin: Git submodules need both an entry in the .gitmodules file and a checked in hash at the location of the submodule. Currently this PR removes the latter.

jordancarlin: I just tried cloning a fresh repo from this PR and it does not have the cvw-arch-verif submodule, so not just an issue with how GitHub is displaying things.

rosethompson: Does the fresh clone contain the submodule in the .gitmodules?

jordancarlin: It's in the .gitmodules file. Upon closer inspection, it seems like the submodule folder has been replaced with an actual folder that contains 4 files.

PR #946, opened 2024-09-05
Fixes missing logic declaration in spi_apb.sv

ross144:
Also updates arty A7 device tree to work correctly with SPI SDC device.


jordancarlin:
Looks like #945 also adds the missing SPI signal in a different place in the file. Doesn't matter which one we go with, but one of the PRs should remove this change.

ross144:
The verilator won't run without the fix.  To avoid the conflict why don't I change this PR so the change is in the same location, then we merge this one.  Hopefully this will avoid a merge conflict when David finishes PR #945.

jordancarlin:
Looks good

PR #952, opened 2024-09-15
Simplified per-config coverage statements

davidharrishmc:


jordancarlin:
Do the same changes need to be made to the rv64gc coverage file?

davidharrishmc:
Yes.  Done!

Issue #841, opened 2024-06-19
Misaligned Faults concurrent with TLB miss


rosethompson:
Misaligned faults may be asserted when a TLB miss occurs.  Because Wally only implements misaligned support for cacheable memory, the address must be translated before cacheablity is known.  Table 15 of the privileged spec defines the priority of faults and Wally places misaligned after page faults. of We need to create a test case and update the HPTW to suppress these faults during a TLB miss.

rosethompson:
Resolved with PR #975

PR #975, opened 2024-09-30
Fixes Issue #841 Misaligned TLB misses

rosethompson:
Checks if the TLB misses and if so suppresses load and store misaligned faults.  Added non-self checking test to verify behavior.  ImperasDV does not match. I believe there may be a different bug either with ImperasDV or our implementation. The test uses PBMT to make a misaligned uncached request with a TLB miss.  The TLB correctly suppresses the fault until the TLB is filled. However, ImperasDV says the load should not cause the fault.


tests/coverage/tlbMisaligned.S
```diff
@@ -0,0 +1,153 @@
+///////////////////////////////////////////
+// tlbGP.S
```
jordancarlin: Header needs to be updated from the file this was copied from

rosethompson: Thank you for finding that.

jordancarlin: Also the test name on line 2

rosethompson: I think everything is fixed now.

PR #982, opened 2024-10-01
fpcalc

stineje:
Improve examples/fp/fpcalc to recognize either 0xValue or xValue to match textbook.  Also improve f128 output to match half/single/double output.


davidharrishmc:
Is it necessary to force the user to write 0x or x?  Can they just write an 8-digit hex number and have that accepted?

stineje:
I was just trying to match the book.  Having it with just a 8-digit hex number is different than the book but do-able.

stineje:
Apologies for the delay; was in a meeting.  Fixed so handles 0x, x, or value.

PR #987, opened 2024-10-02
Removes several unused tests/custom and updates coverage exclusions

rosethompson:
Removes 10 of 16 unused tests from tests/custom.
Adds additional $display so coverage tests print their name as they complete.
Updates code coverage exclusions to not error on PR #986.


jordancarlin:
Also the `custom` list in `tests.vh` probably needs to be updated now.

rosethompson:
Yes to both questions.  Just updated.

PR #994, opened 2024-10-09
Removed axi sdc stuff from the Linux directory.

JacobPease:


jordancarlin:
I believe the Linux Makefile references some of these removed files, so it needs to be updated as well.

JacobPease:
> I believe the Linux Makefile references some of these removed files, so it needs to be updated as well.

Woops. Forgot to add the file. There you go.

PR #995, opened 2024-10-09
Install riscv-isac from riscv-arch-test subdirectory

jordancarlin:


rosethompson:
What is the difference between https://github.com/riscv-software-src/riscv-isac and https://github.com/riscv-non-isa/riscv-arch-test?  I see more recent development on the latter.

jordancarlin:
riscv-isac is one of several tools that make up the RISC-V Architecture Compatibility Tests. Several of them are being merged into the riscv-arch-test repo (which previously only housed the test cases themselves and not the supporting tools).

This updates where we install riscv-isac from to point at the subdirectory in the riscv-arch-test repo since the dedicated riscv-isac repo is being deprecated (will be archived in one week).

PR #1000, opened 2024-10-10
Distributed And-Or MUX

AnonymousVikram:
Worked with @coreyqh

Similar to previous implementation, passing simulation on some, but failing on a lot. Debugging to follow :)

### DIFF
--- a/src/ifu/fetchbuffer.sv
+++ b/src/ifu/fetchbuffer.sv
@@ -29,43 +29,56 @@
 module fetchbuffer import cvw::*; #(parameter cvw_t P) (
 	input  logic        clk, reset,
 	input  logic        StallD, FlushD,
-	input  logic [31:0] writeData,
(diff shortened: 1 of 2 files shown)


src/ifu/fetchbuffer.sv
```diff
@@ -29,43 +29,56 @@
...
+  //     3'b100:  ReadFetchBuffer = Readf2;
+  //     default: ReadFetchBuffer = nop; // just in case?
+  //   endcase
+  //   // issue nop when appropriate
+  //   ReadData = Empty ? nop : ReadFetchBuffer;
+  // end
+
+
+  // Fetch buffer entries anded with read ptr for AO Muxing
+  logic [31:0] DaoArr [2:0];
```
jordancarlin, line 64: Should probably pick a more descriptive name. `DaoArr` is not very clear. Merging anyway for now.

PR #1019, opened 2024-10-17
Installation CI Improvements and README Updates

### LINKED ISSUE #957: Duplicate and conflicting Linux Documentation
The [README](../blob/main/linux/README.MD) file in the linux directory and the [README-linux](../blob/main/docs/README-linux.md) file in the docs directory are redundant and contain conflicting information. The README in the linux directory seems more up to date, so anything not already captured there should be copied over and then the one in the docs directory should be removed.

jordancarlin:
- Fix warnings in installation CI
- Run lint and `wsim rv32i arch32i` in installation CI for distros that fail regression due to Verilator bug to sanity check that the tools mostly installed correctly
- Add installation CI status badge to README
- Document `--no-buildroot` flag of `wally-tool-chain-install.sh`
- Update outdated references to textbook appendices in main README (or would we rather just remove the references entirely? @davidharrishmc)
- Update README formatting of all READMEs
  - Use code blocks that support syntax highlighting
  - Clearly distinguish all code from text explaining the code
  - Use callouts for notes
  - Fix text that was accidentally hidden because it was being interpreted as a comment
  - Fix multiline text that was being interpreted as a single line
- Combine Linux READMEs and update to reflect that the installation script now installs buildroot by default (closes #957)
  - The one piece of the old Linux-README that is not reflected in the new README is below. @rosethompson is this still something that is relevant?

      > ### Generate QEMU linux trace
      > The linux testbench can instruction by instruction compare Wally's committed instructions against QEMU.  To do this QEMU outputs a log file consisting of all instructions executed.  Interrupts are handled by forcing the testbench to generate an interrupt at the same cycle as in QEMU.  Generating this trace will take more than 24 hours.
    >  `cd cvw/linux/testvector-generation`
    > `./genTrace.sh`


rosethompson:
That statement about the QEMU linux trace is no longer relevant. We used to have a QEMU trace we ran against a linux boot. But we no longer support it.

jordancarlin:
> That statement about the QEMU linux trace is no longer relevant. We used to have a QEMU trace we ran against a linux boot. But we no longer support it.

Sounds good. Then I think the documentation should be good to go. We should probably remove that script though because it is still in the linux directory.

rosethompson:
Yes. We should.  Also I'm not sure we need to document the --no-buildroot option in the book.  I think we can just include that in the README or script help.

jordancarlin:
> Yes. We should. Also I'm not sure we need to document the --no-buildroot option in the book. I think we can just include that in the README or script help.

That's what I meant. It's not in the book. I added it to the README in this PR.

rosethompson:
I miss understood.   Unless you want David to review I am happy with this PR.

PR #1021, opened 2024-10-18
Update to GCC 14

### LINKED ISSUE #967: Q not supported in GCC 14
The `wally-toolchain-installation` script pulls the latest version of [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain), which was recently updated to use GCC 14. This fails to recognize `q` as a valid extension in the `-march` string, which causes it to fail when building several of our tests.

jordancarlin:
- Update install script to use the latest GCC (14.2)
- Modify coverage test makefile to manually call the different gcc components so that the lack of Q support in top level GCC can be avoided

Resolves #967


rosethompson:
Interesting It's running the tool install.  Will this happen for all pull requests or just the ones which modify the install script?

jordancarlin:
Only ones that modify the installation script

PR #1023, opened 2024-10-21
Installation and Linux Makefile improvements

jordancarlin:
## CI
- Add custom user level installation to CI
- Ensure installation CI logs are always uploaded if installation fails

## Installation
- Don't suggest looking in the log file if it doesn't exist + log more installation
- Refactor helper functions for increased readability and maintainability
- Update download methods (git and wget/curl) for increased robustness with transient network issues (sometimes experienced by GitHub Actions runners)

## Linux Makefile
- Remove old cleanDriver rule
- Add environment variable validation to ensure setup.sh has been sourced or make is run by installation script (@stineje this address what we were discussing)

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -114,6 +114,7 @@ jobs:
               fi
       # Set environment variables for the rest of the job
       - name: Set Environment Variables
+        if: always()
         run: |
               if [ ! -z ${{ matrix.riscv_path }} ]; then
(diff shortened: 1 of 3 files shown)


jordancarlin:
The CI failure for Ubuntu 20.04 is due to an upstream change that broke compatibility with old GCC unrelated to these changes. I’ve opened a PR to fix it. All other CI tests are passing.

PR #1029, opened 2024-10-24
Changes to nightly regression to get it working with the current repo

slmnemo:


bin/nightly_build.py
```diff
@@ -84,16 +84,20 @@
...
         """
         env_extract_var = 'WALLY'
-        self.base_dir = os.environ.get(env_extract_var)
-        self.base_parent_dir = os.path.dirname(self.base_dir)
+        if os.environ.get(env_extract_var):
+            self.base_dir = os.environ.get(env_extract_var)
+            self.base_parent_dir = os.path.dirname(self.base_dir)
+        else:
+            self.base_dir = basedir
+            self.base_parent_dir = os.path.dirname(self.base_dir)
```
jordancarlin: Won't this cause everything to be cloned into a directory one level above the `basedir` passed as an argument? If we take a directory as an argument it should probably create things in that directory.

slmnemo: Yes, you are correct. Please take a look at the new version.

jordancarlin:
This script really needs a larger refactoring eventually, but the changes look good for now.

PR #1032, opened 2024-10-26
nightly regression-wally fix for buildrootboot and short buildroot squashing, also adding dry-run to regression and jobs to nightly_build.py

slmnemo:


bin/regression-wally
```diff
@@ -363,6 +370,7 @@ os.chdir(regressionDir)
 coveragesim = "questa"  # Questa is required for code/functional coverage
 #defaultsim = "questa"   # Default simulator for all other tests; change to Verilator when flow is ready
 defaultsim = "verilator"   # Default simulator for all other tests
+buildrootbootsim = "questa"
```
jordancarlin: Might be better to define a lockstepsim and use it both for the lockstep buildroot boot and all of the lockstep tests by directory

slmnemo: sure, i can do that.

slmnemo: done

bin/regression-wally
```diff
@@ -526,23 +536,22 @@ def main():
...
     with Pool(processes=min(len(configs),multiprocessing.cpu_count(), ImperasDVLicenseCount)) as pool:
-       num_fail = 0
-       results = {}
-       for config in configs:
-           results[config] = pool.apply_async(run_test_case,(config,))
-       for (config,result) in results.items():
-           try:
-             num_fail+=result.get(timeout=TIMEOUT_DUR)
-           except TimeoutError:
-             pool.terminate()
```
jordancarlin, line 537: Are you deliberately getting rid of the `pool.terminate()` function that you added recently or is this a merge issue?

bin/regression-wally
```diff
@@ -417,17 +426,18 @@ elif (args.fcov):  # only run RV64GC tests on Questa in lockstep in functional c
...
         addTests(tests64gc_fp, sim)
 
 # run derivative configurations and lockstep tests in nightly regression
 if (args.nightly):
-    addLockstepTestsByDir(WALLY+"/tests/coverage", "rv64gc", "questa", 0)
-    addLockstepTestsByDir(WALLY+"/tests/riscof/work/wally-riscv-arch-test/rv64i_m", "rv64gc", "questa", 0)
+    addLockstepTestsByDir(WALLY+"/tests/coverage", "rv64gc", lockstepsim, 0)
+    addLockstepTestsByDir(WALLY+"/tests/riscof/work/wally-riscv-arch-test/rv64i_m", "rv64gc", lockstepsim, 0)
     addTests(derivconfigtests, defaultsim)
+    addTests(bpredtests, defaultsim)
```
jordancarlin: As is, adding the branch predictor tests breaks regression. Several things have changed about how `wsim` works since those were written, so regression will need to be modified if we want those to work. Probably makes sense to add this in a later PR so we can get the rest of this in.

bin/nightly_build.py
```diff
@@ -288,16 +288,17 @@ def execute_makefile(self, makefile_path=None, target=None):
         makefile_location = self.cvw.joinpath(makefile_path)
         os.chdir(makefile_location)
 
-        output_file = self.log_dir.joinpath(f"make-{target}-output.log")
+        if target: 
+            output_file = self.log_dir.joinpath(f"make-{target}-output.log")
+        else: output_file = self.log_dir.joinpath(f"make-output.log")
 
-        command = ["make"]
+        command = ["make", "-j6"]
```
jordancarlin: Any reason to hardcode 6 jobs? For most processes I have been using `--jobs $(nproc --ignore 1)` to use one less thread than the system has. This is more portable to different servers with different hardware.

If we want to use fewer jobs during nightly that is fine, but even then it might make sense to do something like the number of cores divided by 2 or something.

bin/regression-wally
slmnemo: Merge error

bin/regression-wally
slmnemo: Sounds good, I'll comment it out with that tag and mention this to Rose

bin/nightly_build.py
slmnemo: 6 cores was chosen entirely arbitrarily. cores divided by 2 makes more sense. any ideas on how to accomplish that?

jordancarlin: If you want to do cores divided by 2 I think the following would work: `--jobs $(( $(nproc)/2 ))`

jordancarlin: @slmnemo Were you planning to change this too or should I merge?

slmnemo: i'm making the change right now, sorry for the delay.

jordancarlin:
@slmnemo It seems like a rebase went wrong or something and now says that you are recommitting all of the other recent PRs that other people have made.

slmnemo:
A rebase definitely went wrong, will remake PR

PR #1036, opened 2024-10-29
Sv32 Coverage

Huda-10xe:
Renaming VM Coverage and including it in rv32gc coverage file


davidharrishmc:
Can you put the same thing in config/rv32gc, and test that it runs on both configs?

PR #1037, opened 2024-10-29
Added a new spi controller design. Designed as a proof of concept to see if timing issues can be fixed. I intend to work it into existing SPI peripheral.

JacobPease:


rosethompson:
Can you add the SPDX header and your name and date?

JacobPease:
Oh yeah, hold on.

rosethompson:
As I understand, this handles clock generation, sampling SPIIN (pending?), and driving when SPIOUT changed?  Do you think if we paired with is Naichi's fifos's and rest of the controller we could resolve our renaming issues?

rosethompson:
@MikeOpenHWGroup Jacob is getting a ECA failure despite having signed the agreement a long time ago. He's been able to make pull requests and not have issues in the past. We double checked and his email addresses in both commits match what's on file with eclipse.  His eclipse account just says "An unexpected error has occurred."

rosethompson:
image(2)

JacobPease:
> As I understand, this handles clock generation, sampling SPIIN (pending?), and driving when SPIOUT changed? Do you think if we paired with is Naichi's fifos's and rest of the controller we could resolve our renaming issues?

This will require more testing, but from what I've already tested, I think this should pair nicely with the FIFO's. Like you said, it generates the clock, shifting and sampling, and all delays. The different phase and polarity modes should be correct as well. Because of the timing of the signals, this should make sure that the FIFO's aren't desynchronized from the loading of the shift register, and loading the shift register won't cross frame boundaries either.

rosethompson:
@MikeOpenHWGroup Not sure what's going on, but the error went away.

jordancarlin:
@rosethompson I hit the revalidate button on the eclipse site and it seems to be working now. I’ve seen this issue with a few other PRs in the past too.

rosethompson:
I'm going to merge this as it's a good step in the right direction.

PR #1039, opened 2024-10-29
Set MTI_VCO_MODE for all Questa runs, not just with lockstep

### LINKED ISSUE #1038: Questa does not work with ubuntu 24.04 in 32-bit mode
Questa does not run out of the box in Ubuntu 24.04 when launch in the default mode.  By default Questa selects the 32-bit executable and produces the following error.

`/opt/mentor/questasim/bin/../linux/vish: error while loading shared libraries: libXext.so.6: cannot open shared object file: No such file or directory`
 
A work around is to force 64-bit mode with

`MTI_VCO_MODE=64 vsim`

jordancarlin:
Fixes #1038


davidharrishmc:
Do you think it's worth a comment in wsim of what this MTI_VCO_MODE does and why it is necessary?

jordancarlin:
Done

rosethompson:
I wanted to do a quick test to make sure this would haven't a performance impact and it does not.

PR #1050, opened 2024-11-01
Removed SPI hardware interlock test cases

naichewa:
Removed the SPI interlock tests because the new SPI implementation does not pull PREADY low while transmitting/receiving.  Pulling PREADY low hangs up the CPU for long periods of time.  These tests attempted to update the SPI control registers to change clock modes, clock divider, etc during transmission.  This is no longer allowed.


rosethompson:
I've added a comment to your PR so we have documentation as to why these tests are removed.

PR #1057, opened 2024-11-03
Fixed buildroot lockstep, removed source.sh dependency in nightly build

slmnemo:


config/buildroot/imperas.ic
jordancarlin, line 1: Doesn't this cause some of the buildroot configuration files to be in `/config` and some to be in `/config/deriv`? Seems like it would be better to keep all of the files together. Maybe `derivgen.pl` could create the link/copy `imperas.ic` for the buildroot configuration instead. @davidharrishmc thoughts?

slmnemo: I tried this, I was planning on implementing it this way. imperas.ic is not sourced if it is found under `config/deriv`. This would require a change to the testbench and I cannot find where imperas.ic is loaded into the testbench. This also requires a change in the perl, but that's like two more lines.

jordancarlin:
It’s in wsim. Part of the prefix.

slmnemo:
Ah, do we have a lookup for whether a config is a derivative config? Do we want such a thing in wsim? We can use that to run derivative configs and I can add a linker script to the perl then and revert some of the changes made above.

slmnemo:
> Ah, do we have a lookup for whether a config is a derivative config? Do we want such a thing in wsim? We can use that to run derivative configs in lockstep and I can add a linker line to the perl then and revert the existing symlink.

Could also just say "if it doesn't exist, look in `config/deriv` for the correct imperas.ic. How do you feel about that @jordancarlin?

slmnemo:
Running a lockstep run right now, but this should work now.

jordancarlin:
The concern with that is not all derived configs will use the same imperas.ic as their base config. As features get enabled/disabled they need different imperas configurations. See the recent messages in the regression-testing channel of slack. For now it would probably be better to only copy/link for the buildroot config.

bin/wsim
```diff
@@ -91,6 +91,8 @@ prefix = ""
 if (args.lockstep or args.lockstepverbose or args.fcov or args.fcovimp):
     if (args.sim == "questa" or args.sim == "vcs"):
         prefix = "IMPERAS_TOOLS=" + WALLY + "/config/"+args.config+"/imperas.ic"
+        if not os.path.isfile(prefix): # If config is a derivative, look for imperas.ic in derivative configs
+            prefix = "IMPERAS_TOOLS=" + os.path.join(WALLY, "config", "deriv", args.config, "imperas.ic")
```
jordancarlin, line 95: Should probably use the same format for these paths for both the deriv and not deriv versions. I don't think it matters which, but probably better not two have 2 different versions of the same thing back to back.

PR #1058, opened 2024-11-03
Fixed rmm rounding mode bug

### LINKED ISSUE #1056: Fround.h with rmm rounding mode mismatched with ImperasDV
When running fround.h tests with explicit rmm rounding, Imperas mismatches. It appears that Wally rounds correctly, just to the wrong value. 

For example, Wally rounds 0xc872 (decimal -8.890625) to 0xc800 (decimal -8) while the reference rounds to 0xc880 (decimal -9). 

```
$ wsim rv32gc tests/rv32/ZfaZfh/WALLY-COV-fround.h.elf --fcov 

# Info (IDV) Instruction executed prior to mismatch '0x80000e54(rvtest_entry_point+e54): 4445ce53 fround.h f28,f11,rmm'
# Error (IDV) FPR register value mismatch (HartId:0, PC:0x80000e54 rvtest_entry_point+e54):
# Error (IDV) Mismatch 0> FPR f28
# Error (IDV)   . dut:0xffffffffffffc800
# Error (IDV)   . ref:0xffffffffffffc880
# Error (IDV) testbench.idv_trace2api.state_compare @ 22340: MISMATCH
```

The assembly file leading to mismatch is here 
[WALLY-COV-fround.h.S.tar.gz](https://github.com/user-attachments/files/17611950/WALLY-COV-fround.h.S.tar.gz)

coreyqh:
Rounding logic case statement had 3'b101 (a reserved rounding mode value) associated with the rmm rounding mode instead of 3'b100. I changed this case to the correct value and fixed the ImperasDV mismatch for fround with explicit rmm rounding mode

close #1056

### DIFF
--- a/src/fpu/fround.sv
+++ b/src/fpu/fround.sv
@@ -126,7 +126,7 @@ module fround import cvw::*;  #(parameter cvw_t P) (
       3'b001:  RoundUp = 0;               // RZ
       3'b010:  RoundUp = Xs & (Rp | Tp);  // RN
       3'b011:  RoundUp = ~Xs & (Rp | Tp); // RP
-      3'b101:  RoundUp = Rp;              // RNTA
+      3'b100:  RoundUp = Rp;              // RNTA
       default: RoundUp = 0;               // should never happen
     endcase
 


src/fpu/fround.sv
```diff
@@ -126,7 +126,7 @@ module fround import cvw::*;  #(parameter cvw_t P) (
       3'b001:  RoundUp = 0;               // RZ
       3'b010:  RoundUp = Xs & (Rp | Tp);  // RN
       3'b011:  RoundUp = ~Xs & (Rp | Tp); // RP
-      3'b101:  RoundUp = Rp;              // RNTA
+      3'b100:  RoundUp = Rp;              // RNTA
```
jordancarlin, line 129: Should this comment say `RMM`? What's RNTA?

coreyqh: RNTA (round to nearest, ties away) is the IEEE nomenclature for the rmm (round to nearest, ties to max magnitude) rounding mode according to the risc-v SoC book. I didn't think to change the comment since they mean the same thing. Should I change?

jordancarlin:
Should probably be consistent with the rest of the code base. Do you know what we use everywhere else when referring to this rounding mode?

src/fpu/fround.sv
jordancarlin: Should probably be consistent with the rest of the code base. Do you know what we use everywhere else when referring to this rounding mode?

jordancarlin:
Going to merge for now but we can look at the comments later

PR #1059, opened 2024-11-04
Fetch buffer passes regression!

jordancarlin:
Update hazards and reset value so fetch_buffer passes regression.

### DIFF
--- a/sim/questa/wave.do
+++ b/sim/questa/wave.do
@@ -6,43 +6,43 @@ add wave -noupdate /testbench/reset
 add wave -noupdate /testbench/memfilename
 add wave -noupdate /testbench/dut/core/SATP_REGW
 add wave -noupdate /testbench/dut/core/InstrValidM
-add wave -noupdate -group HDU -expand -group hazards /testbench/dut/core/hzu/RetM
-add wave -noupdate -group HDU -expand -group hazards -color Pink /testbench/dut/core/hzu/TrapM
-add wave -noupdate -group HDU -expand -group hazards /testbench/dut/core/ieu/c/LoadStallD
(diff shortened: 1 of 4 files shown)


davidharrishmc:
Is this supposed to change 159 files?

jordancarlin:
It is also updating the fetch buffer branch with all of the changes from the main branch, so yes. If you'd rather I can do that as a separate PR.

The actual changes are all in the last commit if you want to look at them.

davidharrishmc:
Let's do a review at the research group meeting.

jordancarlin:
Sounds good. I will open another PR in the meantime to update the fetch_buffer branch so this is easier to review.

rosethompson:
I don't think I can make it to your review later today, but I have some comments about the fetch buffer.
1. I like very simple design of the fetch buffer and how it decouples the fetch stage stalls from the rest of the pipeline.
2. Is there a way to simplify the spill logic? Specifically it looks like the spills are unchanged so a 4 byte spill across two cache lines will still cause a spill delay.  Are we still paying the extra cycle of latency? If so how difficult will it be to move the spill logic after the fetch buffer so it can absorb the cycle?
I am very pleased with the simplicity of the design.

jordancarlin:
Thanks @rosethompson.

The goal is definitely to modify the spill logic. We wanted to get the minimal viable version working first and then begin optimizing from there. Right now there is no performance benefit from the fetch buffer. Avoiding the latency for spills if the second instruction is already in the buffer is one of the big optimizations we want to make. Several other stalls can also probably be eliminated.

rosethompson:
I thought that might be the case. I approve of these changes and I look forward to the future!

PR #1062, opened 2024-11-04
Remove unused file from Linux/testvector-generation

rosethompson:
Removes the majority of files from the testvector generation directory.  We probably consolidate further by replacing fixBinMem with objcopy and then move genInitMem.sh into the linux directory.

I also modified the regression build scripts to parallelize the objdump and memfile generation.


tests/riscof/Makefile
```diff
@@ -31,7 +31,7 @@ arch%: config$$*.ini | $(work_dir) $(arch_workdir)
 # Generate wally-riscv-arch-test targets
 wally%: config$$*.ini | $(work_dir) $(wally_workdir)
 	riscof run --work-dir=$(work_dir)/$@ --config=$< --suite=$(wally_dir)/riscv-test-suite/ --env=$(wally_dir)/riscv-test-suite/env --no-browser --no-dut-run
-	$(MAKE) -f makefile-memfile WORKDIR=$(work_dir)/$@
+	$(MAKE) -f makefile-memfile --jobs WORKDIR=$(work_dir)/$@
```
jordancarlin: Ideally make is called using the `--jobs` or `--jobs #` flag which will then let it run all sub makes using that job server. Adding `--jobs` inside the Makefile can cause some weird issues with resetting the job count that it uses which overwhelms certain servers because it is already running many different Makefiles (for example when `make --jobs` is run at the top-level)

rosethompson:
Ok I didn't know that. I've reverted that commit.

jordancarlin:
Sounds good. If you run make using `make --jobs` you should still get the benefits of it running that step in parallel.

PR #1064, opened 2024-11-05
Added SCKMODE 10 and 11 delay cases to regression tests

naichewa:


jordancarlin:
Looks like several submodules are being added/reverted. I assume this isn’t intentional.

Issue #1053, opened 2024-11-02
Remove deprecated Linux scripts


jordancarlin:
As mentioned in https://github.com/openhwgroup/cvw/pull/1019#issuecomment-2423552575, several of the scripts in the `linux/testvector-generation` directory are no longer supported. At a minimum this includes `genTrace.sh`, but I suspect there are others that are no longer used as well.

rosethompson:
Looks like we only use fixBinMem and that's just to reverse the byte order of the QEMU memory dump.

jordancarlin:
Mostly resolved with #1062. Still the potential to 
> consolidate further by replacing fixBinMem with objcopy and then move genInitMem.sh into the linux directory.

davidharrishmc:
Did this one get resolved?

jordancarlin:
Yes. Fixed with #1062 and #1081

PR #1081, opened 2024-11-10
Use objcopy to fix linux testvector byte order

jordancarlin:
- Remove fixBinMem.c and use objcopy instead (as discussed in #1053) @rosethompson 
- Also remove seemingly unused bootmem.txt file. I believe a newer version of this is built and used when running make


davidharrishmc:
I think bootmem.txt is needed by synthesis to avoid optimizing away the
IROM. We presumably want it to not change. Is there any chance this file is
used for that and shouldn’t depend on make?

jordancarlin:
The rom1p1r.sv file preloads the rom with `fpga/src/boot.mem`, not `linux/bootmem.txt`. I couldn't find anywhere that this was used. I could be wrong though.

```verilog
    if (PRELOAD_ENABLED) begin
      if (DATA_WIDTH == 64) begin
        `ifdef VERILATOR
            // because Verilator doesn't automatically accept $WALLY from shell
            string       WALLY_DIR = getenvval("WALLY"); 
            $readmemh({WALLY_DIR,"/fpga/src/boot.mem"}, ROM, 0);  // load boot ROM for FPGA
        `else
            $readmemh({"$WALLY/fpga/src/boot.mem"}, ROM, 0);  // load boot ROM for FPGA
        `endif
      end else begin // put something in the ROM so it is not optimized away
        ROM[0] = 'h00002197;
      end
    end
```

rosethompson:
This bootmem.txt file looks like part of an objdump for the QEMU bootloader. However the buildroot simulation gets it's startup from linux-testvectors/bootmem.bin.   The fpga gets it's zsbl from the files Jordan mentioned.  The IROM is preloaded with 

```
      end else begin // put something in the ROM so it is not optimized away
        ROM[0] = 'h00002197;
      end

```

for ASIC synth.

There are no references to linux/bootmem.txt in any other files.

PR #1088, opened 2024-11-12
Fixes lint warnings in loggers.sv updates spi device tree for vcu108

rosethompson:
Worked with Kaitlin to fix the lint warnings preventing her from making progress from including benchmarks in the nightly regression tests.
Updated spi clock frequency so it works correctly with vcu108.


fpga/src/fpgaTop.sv
```diff
@@ -216,11 +216,8 @@ module fpgaTop
                     .HADDR, .HWDATA, .HWSTRB, .HWRITE, .HSIZE, .HBURST, .HPROT,
                     .HTRANS, .HMASTLOCK, .HREADY, .TIMECLK(1'b0), 
                     .GPIOIN, .GPIOOUT, .GPIOEN,
-                    .UARTSin, .UARTSout, .SDCIn, .SDCCmd, .SDCCS(SDCCSin), .SDCCLK(SDCCLKInternal), .ExternalStall(RVVIStall));
+                    .UARTSin, .UARTSout, .SDCIn, .SDCCmd, .SDCCS(SDCCSin), .SDCCLK(SDCCLK), .ExternalStall(RVVIStall));
```
jordancarlin: Should this just use `.SDCCLK` like all the other signals with the same name?

rosethompson:
Yes

rosethompson:
Well this is interesting. We tested this with verilator but not questa and questa is not happy with loggers.sv.  I'm going to convert to a draft and fix.

jordancarlin:
Strange. What is Verilator complaining about in the original version? We've been able to get away using bitwise operators everywhere else so it would also be nice to stick with that if possible instead of switching to logical operators. Was Verilator complaining about them otherwise?

rosethompson:
Questa gave an error about TEST not being constant inside a generate which makes sense because it's a plus arg.  Verilator seems to be ok with this.  I've refactored the code to avoid the generate.

I removed the logic operators.  Kaitlin and I were having issues with the bitwise operators throwing bitwidth warnings with Verilator, but after the other issues have been fixed these seem to have gone away as well.

rosethompson:
@jordancarlin Are you happy with the changes? If so I'd like to get this merged so @slmnemo can work on integrating performance monitoring in the nightly regression.

jordancarlin (approved):

Issue #909, opened 2024-08-11
CacheSim Mismatches


jordancarlin:
Now that rv64gcCacheSim.py is working again (#908), several of the tests report cache mismatches. The problematic ones are listed below. All of the other tests match.

```
Commencing test coverage64gc:
wsim --sim questa rv64gc coverage64gc --params "I_CACHE_ADDR_LOGGER=1\'b1 D_CACHE_ADDR_LOGGER=1\'b1" > /dev/null
Running the ICache simulator.
SUCCESS! There were no mismatches between Wally and the sim.
Running the DCache simulator.
Result mismatch at address 00000080010000. Wally: E, Sim: H
Result mismatch at address 00000080011010. Wally: E, Sim: H
Result mismatch at address 00000080012020. Wally: E, Sim: H
Result mismatch at address 00000080015008. Wally: D, Sim: H
Result mismatch at address 00000080010000. Wally: E, Sim: H
Result mismatch at address 00000080011010. Wally: E, Sim: H
Result mismatch at address 00000080012000. Wally: E, Sim: H

Commencing test arch64i:
wsim --sim questa rv64gc arch64i --params "I_CACHE_ADDR_LOGGER=1\'b1 D_CACHE_ADDR_LOGGER=1\'b1" > /dev/null
Running the ICache simulator.
Result mismatch at address 00000080030fbc. Wally: E, Sim: H
Result mismatch at address 00000080026970. Wally: E, Sim: H
Result mismatch at address 00000080022db8. Wally: H, Sim: E
Result mismatch at address 0000008002aaf8. Wally: E, Sim: H
Result mismatch at address 000000800301e4. Wally: E, Sim: H
Running the DCache simulator.
SUCCESS! There were no mismatches between Wally and the sim.

Commencing test wally64priv:
wsim --sim questa rv64gc wally64priv --params "I_CACHE_ADDR_LOGGER=1\'b1 D_CACHE_ADDR_LOGGER=1\'b1" > /dev/null
Running the ICache simulator.
SUCCESS! There were no mismatches between Wally and the sim.
Running the DCache simulator.
Result mismatch at address 00000080003380. Wally: D, Sim: E
Result mismatch at address 00000080004380. Wally: D, Sim: E
Result mismatch at address 00000080005380. Wally: D, Sim: E
Result mismatch at address 00000080006380. Wally: D, Sim: E
Result mismatch at address 00000080003180. Wally: M, Sim: H
[...]

PR #1090, opened 2024-11-13
Fixes multiple cache bugs and CacheSim.py bugs

rosethompson:
Resolves issue #909. Please merge after PR #1088 is merged.

To summarize there were two main bugs in the cache.
1. A store hit sets the dirt bit.  If the following instruction reads the same cache set, this dirty bit was not read as dirty but rather as cleared because of a structural hazard.  The original cache design assumed the pipeline would insert a pipeline stall to avoid the hazard. The solution was a simple forward of the write data to the read data when writing the dirty bit.  This hazard would cause an issue when the second instruction would need to cause a writeback.  The cbo.clean and cbo.flush instructions revealed the bug.
2. The LRU was not correctly updated when the cache was stalled.  The wrong set address was sent the LRU.

The cache simulator did not support cbo instructions and now does.


davidharrishmc:
Should there be variable names and/or comments in cache.sv to make it clearer that forwarding it taking place?  The 2 is a bit obscure.

jordancarlin:
Is there a reason the default sim is being changed from verilator to questa?

rosethompson:
I do not see the change for the default sim in the PR.

rosethompson:
Oh wait you're talking about the cache simulator. I can change that back. It was faster to run questa while I was working on the bug rather than use verilator.

sim/rv64gc_CacheSim.py
```diff
@@ -63,10 +65,11 @@ def main():
     parser = argparse.ArgumentParser(description="Runs the cache simulator on all rv64gc test suites")
     parser.add_argument('-p', "--perf", action='store_true', help="Report hit/miss ratio")
     parser.add_argument('-d', "--dist", action='store_true', help="Report distribution of operations")
-    parser.add_argument('-s', "--sim", help="Simulator", choices=["questa", "verilator", "vcs"], default="verilator")
+    parser.add_argument('-s', "--sim", help="Simulator", choices=["questa", "verilator", "vcs"], default="questa")
```
jordancarlin: @rosethompson right here

PR #1091, opened 2024-11-14
Fix RVVI for RV32

jordancarlin:
- RV32 only CSRs set up correctly
- PMP CSRs updated to scale for RV32 or RV64
- Also updated the formatting of wallyTracer.sv because I was struggling to figure out what was happening with it in its current form. Recommend reviewing with view whitespace changes turned off because of this.


davidharrishmc:
We don’t yet use mseccfgh (0x757), but OSU is working on Zkr randomness, which will use it.  Might as well add it alongside menvcfgh while we are thinking about this.

jordancarlin:
Until that register actually exists there is nothing to tie it to. If I point it at a nonexistent register it generates errors. I can just tie it to 0 for now and not fully hook it up, but at that point it might make sense to just wait for it to be implemented.

davidharrishmc:
It’s enough places to connect that I think it might be worth doing now while you are familiar, and tying to 0, with a comment on the line that changes when the register exists.

jordancarlin:
Added mseccfgh and rearranged all of the CSRs to match the way they are laid out in table I.9

PR #1094, opened 2024-11-14
added Zicsr (32 and 64) to run in Imperas

ahlyssa:


davidharrishmc:
Please make a concurrent cvw-arch-verif PR to create the tests so they can be found when you run them.

PR #1103, opened 2024-11-16
Made minor changes to the controller to clean up the logic. Still need to simplify the first always block.

JacobPease:


JacobPease:
Weird. I just changed this file after pulling. I'll look into it.

jordancarlin:
Pretty sure it was a lint_off issue that I fixed. Should be an easy resolution.

jordancarlin:
You seem to have accidentally reverted other recent PRs when fixing the merge conflict...

rosethompson:
We should try to get this merged. I'm a bit confused by the commit history. It shows many extra files changed?  Is the best way to merge to swash into a single commit so we don't have the unnecessary changes in the history?

jordancarlin:
I only see 2 files being changed...

rosethompson:
It's more the history has a bunch of files bunch of file changed. Maybe that's ok.

jordancarlin:
I think those were all force pushed away. Looking through the 3 commits that are actually included in the PR I don't see any extra files being modified. If we're worried about it though then squash merging should be fine. @JacobPease would just need to hard reset his branch afterwards.

rosethompson:
I think we can just merge the pr.

Issue #1092, opened 2024-11-14
WALLY-lrsc fails in RV32 lockstep


jordancarlin:
The wally-riscv-arch-test lrsc test is mismatching with ImperasDV for rv32gc.

```
$  wsim rv32gc tests/riscof/work/wally-riscv-arch-test/rv32i_m/privilege/src/WALLY-lrsc-01.S/ref/ref.elf --sim questa --lockstep

# Info (IDV) ---------------------------------------------------
# Info (IDV) Instruction executed prior to mismatch '0x8000003c(rvtest_entry_point+3c): 18bfa62f sc.w    x12,x11,(x31)'
# Error (IDV) GPR register value mismatch (HartId:0, PC:0x8000003c rvtest_entry_point+3c):
# Error (IDV) Mismatch 0> GPR x12
# Error (IDV)   . dut:0x00000001
# Error (IDV)   . ref:0x00000000
# Error (IDV) testbench.idv_trace2api.state_compare @ 1150: MISMATCH
```

PR #1104, opened 2024-11-16
Privileged functional coverage tests run in regression-wally --fcov

davidharrishmc:
Also fixed initialization of mtimecmp

### DIFF
--- a/bin/regression-wally
+++ b/bin/regression-wally
@@ -301,7 +301,8 @@ def addLockstepTestsByDir(dir, config, sim, fcovMode):
         for dirpath, dirnames, filenames in os.walk(os.path.abspath(dir)):
             for file in filenames:
                 # fcov lockstep only runs on WALLY-COV-ALL.elf files; other lockstep runs on all files
-                if (file.endswith(".elf") and fcovMode == 0 or file.endswith("ALL.elf") and fcovMode == 1):
+                if ((file.endswith(".elf") and (fcovMode == 0 or "tests/priv" in dir)) or 
+                    (file.endswith("ALL.elf") and fcovMode == 1)):
                     fullfile = os.path.join(dirpath, file)
                     fields = fullfile.rsplit('/', 3)
                     if (fields[2] == "ref"):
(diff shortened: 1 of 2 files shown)


bin/regression-wally
```diff
@@ -419,9 +420,11 @@ if (args.ccov):  # only run RV64GC tests on Questa in code coverage mode
     addTests(tests64gc_nofp, coveragesim)
     if (args.fp):
         addTests(tests64gc_fp, coveragesim)
-elif (args.fcov):  # only run RV64GC tests on Questa in lockstep in functional coverage mode
+elif (args.fcov):  # run tests in lockstep in functional coverage mode
     addLockstepTestsByDir(WALLY+"/addins/cvw-arch-verif/tests/rv32/", "rv32gc", coveragesim, 1)
     addLockstepTestsByDir(WALLY+"/addins/cvw-arch-verif/tests/rv64/", "rv64gc", coveragesim, 1)
+    addLockstepTestsByDir(WALLY+"/addins/cvw-arch-verif/tests/priv/rv32/", "rv32gc", coveragesim, 1)
```
jordancarlin, line 426: Doesn’t this break regression currently because of #1092?

bin/regression-wally
```diff
@@ -301,7 +301,8 @@ def addLockstepTestsByDir(dir, config, sim, fcovMode):
         for dirpath, dirnames, filenames in os.walk(os.path.abspath(dir)):
             for file in filenames:
                 # fcov lockstep only runs on WALLY-COV-ALL.elf files; other lockstep runs on all files
-                if (file.endswith(".elf") and fcovMode == 0 or file.endswith("ALL.elf") and fcovMode == 1):
+                if ((file.endswith(".elf") and (fcovMode == 0 or "tests/priv" in dir)) or
```
jordancarlin, line 304: I know UET want to use some of the riscv-arch-tests for functional coverage. If you’re changing this anyway does it make sense to generalize so any of the riscof tests can be run with fcov instead of just priv tests?

jordancarlin: Oh. Wait. This is referring to tests/priv in cvw-arch-verif. I misunderstood. Seems fine then.

bin/regression-wally
jordancarlin: Oh. Wait. This is referring to tests/priv in cvw-arch-verif. I misunderstood. Seems fine then.

PR #1115, opened 2024-11-17
Updated nightly instructions in readme to use curl

slmnemo:


davidharrishmc:
The nightly_build URL for me seems to be:
https://github.com/openhwgroup/cvw/blob/main/bin/nightly_build.py

Is the githubusercontent URL in the curl statement exactly equivalent?

davidharrishmc:
Would you like to add you and Jordan to the email?

slmnemo:
I only included two to show how to format with a real example. The githubusercontent URL should be equivalent for actually grabbing the raw text in the script. The link you provided is not going to work because the github site does not curl correctly from my testing from what I remember

PR #1141, opened 2024-11-26
Add support for SUSE family distros (tested on OpenSUSE) to installation script

jordancarlin:
The one major family of Linux distros that still wasn't supported was SUSE. This adds support for the latest version of OpenSUSE Leap (and should be compatible with SLES as well). Besides RHEL, this is the other major distro supported by EDA tool vendors, so it seemed worthwhile to include. Especially because it was very little effort to add after the recent refactoring of the scripts.

I also took the chance to do additional refactoring of the installation script for improved maintenance and readability.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -73,6 +73,11 @@ jobs:
             os: ubuntu-latest
             image: almalinux:9
             imageFamily: redhat
+          # SUSE Installations
+          - name: opensuse-15.6
+            os: ubuntu-latest
+            image: opensuse/leap:15.6
+            imageFamily: suse
           # User level installation
           - name: user-install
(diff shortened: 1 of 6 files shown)


jordancarlin:
All of the installations actually succeeded. The failures were all caused by Verilator failing regression with the updated Verilog Ethernet suite from #1135

jordancarlin:
Hmm. Strange that SUSE is failing. It worked when I tested it locally. Will investigate.

jordancarlin:
Oh. It’s downloading the site-setup.sh script from the main repository, which doesn’t have the changes needed to support SUSE that are included in this PR. The installation was successful but the path was configured wrong for regression because of that. 

Should be good to merge.

PR #1157, opened 2024-11-30
List compressed instructions in objdump (also disables pseudo instructions)

jordancarlin:
Add `-Mno-aliases` to all objdump commands. This makes compressed instructions come through correctly in the objdump files (`c.lw` instead of `lw`, `c.addi16sp` instead of jus `addi`, etc.). It does have the downside of also not using any pseudoinstructions, so `csrr` now comes through as `csrrw zero`. I think it's still worth it, but could go either way.

### DIFF
--- a/benchmarks/coremark/Makefile
+++ b/benchmarks/coremark/Makefile
@@ -23,7 +23,7 @@ PORT_CFLAGS = -g -mabi=$(ABI) -march=$(ARCH) -static -falign-functions=16 \
 	-fno-delete-null-pointer-checks -fno-rename-registers --param=loop-max-datarefs-for-datadeps=0 \
 	-funroll-all-loops --param=uninlined-function-insns=8 -fno-tree-vrp -fwrapv -fipa-pta \
 	-nostdlib -nostartfiles -ffreestanding -mstrict-align \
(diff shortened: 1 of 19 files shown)


davidharrishmc:
I'm wary about changing the examples, because we won't be able to change the book at this point, and I don't want to add more compiler flags than necessary because it complicates Chapter 3. 

Is this solving a problem that's affecting you in some other way?

jordancarlin:
Was debugging a test failure and the objdumps currently differ from Imperas and Breker in this regard. Seemed like it would be easier for all three versions to align (briefly confused me when I was seeing different instructions in different places).

Can definitely leave the examples alone and if you’d rather just skip this entirely I think that’s fine too. Just figured I’d at least mention it now that we’ve switched imperas over to output instructions without pseudoinstructions and with the c. prefix.

davidharrishmc:
How about turn on the flag for specific things you need?  Maybe for objdumps of Breker code?  Leave historic stuff the way it was?

jordancarlin:
Sounds good. Do you want it turned on for cvw-arch-verif so that it matches with Imperas?

davidharrishmc:
I don’t have a strong need for that, but you could tell the team and they could turn it on if they feel they need it when debugging.  Vikram and Corey are the only ones who have found any bugs yet and needed the objdump.

PR #1160, opened 2024-12-01
Support elf files on all simulators

jordancarlin:
Add support for elf files on Verilator and update VCS to be in line with the other simulators

### DIFF
--- a/bin/wsim
+++ b/bin/wsim
@@ -64,6 +64,9 @@ if(args.testsuite.endswith('.elf') and args.elf == ""): # No --elf argument; che
         print("ELF file not found: " + args.testsuite)
         exit(1)
 
+if (ElfFile != ""):
+    args.args += " " + ElfFile
+
 if(args.lockstep and not args.testsuite.endswith('.elf') and not args.testsuite == "buildroot"):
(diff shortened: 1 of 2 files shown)


davidharrishmc:
Very nice!

This works when run on a .elf.  It's good enough for @AishwaryaPrabhu1 to move ahead.

If I try to run it with --elf <file> with no .elf suffix, the testbench doesn't make the memfile properly.  I don't love the --elf option and having to give a bogus test suite.  Dou think there is a better factoring?

```
wsim --sim verilator rv64gc null --elf gpio
...
make[1]: Entering directory '/home/harris/cvw/sim/verilator'
make[1]: *** No rule to make target '/home/harris/cvw/examples/C/gpio/gpio.memfile'.  Stop.
make[1]: Leaving directory '/home/harris/cvw/sim/verilator'
Error: Could not open file /home/harris/cvw/examples/C/gpio/gpio.memfile
```

jordancarlin:
I’ll look into it. I thought --elf was supposed to be for when the file itself didn’t have a .elf extension.

jordancarlin:
And I suspect the same issue occurs with any of the other simulators. I think the Makefile the testbench calls is only set up to produce memfiles from .elf files.

PR #1162, opened 2024-12-02
wsim refactoring + better support for elfs

jordancarlin:
- Refactor wsim into smaller, more modular functions for easier maintenance and better clarity on what is specific to each simulator and what is common to them all.
  - Also a step towards allowing for tab completion of wsim commands.
- Update string concatenation to use f-strings for increased readability and faster performance (also is the recommended way to do strings in Python).
- Update Verilator `make` command to use the system make instead of hardcoding a specific path (important if a newer version of make is installed on a system)
- Make the positional testsuite argument optional if you are passing an elf with the `--elf` flag. This allows for commands like `wsim rv64gc --elf hello` and avoids the need for dummy test suites.
- Update the Makefile used by the testbench to support creating memfiles even if the elf file does not with a `.elf` extension.

### DIFF
--- a/bin/wsim
+++ b/bin/wsim
@@ -20,7 +20,7 @@ WALLY = os.environ.get('WALLY')
 def parseArgs():
     parser = argparse.ArgumentParser()
     parser.add_argument("config", help="Configuration file")
-    parser.add_argument("testsuite", help="Test suite or path to .elf file")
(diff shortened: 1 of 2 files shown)


jordancarlin:
I recommend reviewing the diff with "hide whitespace" turned on. Otherwise it makes it look like I changed significantly more than I did because of indentation.

PR #1173, opened 2024-12-03
Adding another signal for VM Coverage

Huda-10xe:
To cover all Page levels this signal is need in VM Coverage


Huda-10xe:
@rosethompson could you review this tracer to let me know if I've extracted the signals from the correct stages in the pipeline and flopped them properly. I've added comments to help explain what each signal is supposed to be used for. If there's further clarification required let me know.

rosethompson:
This looks correct. I'll merge now.

PR #1182, opened 2024-12-09
Add --define to wsim

jordancarlin:
- Add `--define` flag to `wsim` that takes arguments of the standard form `"+define+USE_IMPERAS_DV"` to pass as macro definition overrides to simulators. Support added to all 3 individual simulator scripts.
- Refactor lots of the lockstep and fcov logic to be in the wsim level for easier maintainability across simulators.


jordancarlin:
One thing to think about here. 

I went with the syntax of `--define "+define+USE_IMPERAS_DV"` since that is in line with the standard way of defining macros at the command line for simulators. Since this is already behind a `--define` flag, it is somewhat redundant. We could have it be something like `--define "USE_IMPERAS_DV"` instead to simplify it, but I don't know how far we want to go into defining our own syntax. We were already forced to come up with our own syntax for `--params` since the simulators all handle that differently, but there is consistency with `+define` so I'd be inclined to stick what that even if it is redundant and more verbose.

davidharrishmc:
Simple is great.

PR #1188, opened 2024-12-10
Initial attempt at Ubuntu derivative distro support, focusing on Linux Mint

jordancarlin:
Update version detection logic to handle Ubuntu derivatives. Installation is still known to be failing on Linux Mint 22 because of other issues I'm still troubleshooting. Pushing this for now because it may solve the issues in Linux Mint 21 without further changes.

I'm not going to list Ubuntu derivatives as officially supported by the installation script because we won't be testing on all of them, but if we can make it work most of the time that would still be great.

### DIFF
--- a/bin/wally-distro-check.sh
+++ b/bin/wally-distro-check.sh
@@ -76,10 +76,28 @@ if [[ "$ID" == rhel || "$ID_LIKE" == *rhel* ]]; then
 elif [[ "$ID" == ubuntu || "$ID_LIKE" == *ubuntu* ]]; then
     export FAMILY=ubuntu
     if [ "$ID" != ubuntu ]; then
-        printf "${WARNING_COLOR}%s%s\n${ENDC}" "For Ubuntu family distros, the Wally installation script has only been tested on standard Ubuntu. Your distro " \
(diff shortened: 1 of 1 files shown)


jordancarlin:
Looks like something happened related to authentication with the Red Hat builds. Traced it back to an updated version of PAM. It's unrelated to this change, so assuming the other distros pass CI we can merge this. I'll investigate that separately and open another PR with a workaround when I figure something out.

PietDelaney (approved):

PietDelaney:
Looks good using /etc/os-release:

    UBUNTU_CODENAME=jammy

for Linux Mint 21.3

With Gerrit its easy to checkout a proposed commit. I haven't figured out how to do it with this github code review framework.

jordancarlin:
@PietDelaney thanks for the confirmation regarding the `UBUNTU_CODENAME` on 21.3.

For future reference, if you have the GitHub CLI installed (gh) you can checkout a PR by running `gh pr checkout $ID`. Otherwise you can run the following sequence of git commands:
```
git fetch origin pull/$ID/head:$BRANCHNAME
git checkout $BRANCHNAME
```
where $ID is the PR number and $BRANCHNAME is the name of the local branch you want to create. If the OpenHWGroup repo is not the origin repo (which might be the case if you cloned a fork as the origin) then change origin accordingly (likely to upstream).

PR #1194, opened 2024-12-12
fixed typos in intdiv regression script; added delay state in fp-testbench fsm to fix a bug

kevindkim723:
no RTL changes; all intdiv configs pass divremsqrt tests


davidharrishmc:
Good to have this patch.  Does this by chance fix Verilator simulation of testfloat for the main branch too?

PR #1199, opened 2024-12-19
Breker test support

jordancarlin:


davidharrishmc:
The breker stuff may confuse a reader because it isn't mentioned in the book, and will be inaccessible to our team as well after May (when the license expires).  Could you add a README in tests/Breker saying what this stuff is about and that a license is needed (and not generally available to noncommercial users)?  And, where it might help in the other files such as wally.do and testbench.sv, add a comment line about what this is for and that a license is required to use it?

jordancarlin:
Updated the documentation.

PR #1200, opened 2024-12-20
Updating WallyTracer for VM signals

Huda-10xe:
All the MMU signals were being propagated from the Memory to WB stage regardless of the data coming from IMMU or DMMU. This has now been changed such tat ITLB data is propagated from the IF stage to the WB stage. Also the pipeline stages that interact with the HPTW are propagating data whenever the HPTW is active i.e dependent on the SelHPTW signal.

With these changes once riscv-arch-test were run the following coverage report was generated
Screenshot from 2024-12-20 20-13-53

### DIFF
--- a/testbench/common/wallyTracer.sv
+++ b/testbench/common/wallyTracer.sv
@@ -45,6 +45,7 @@ module wallyTracer import cvw::*; #(parameter cvw_t P) (rvviTrace rvvi);
   logic                  InstrValidM, InstrValidW;
   logic                  StallE, StallM, StallW;
   logic                  GatedStallW;
+  logic                  SelHPTW;
   logic                  FlushD, FlushE, FlushM, FlushW;
   logic                  TrapM, TrapW;
   logic                  HaltM, HaltW;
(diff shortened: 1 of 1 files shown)


rosethompson:
Why is the register enable between the Fetch and Decode stage SelHPTW and not ~StallD?  For example
`flopenrc #(P.XLEN) VAdr_iDReg (clk, reset, 1'b0, SelHPTW, VAdr_iF, VAdr_iD); //Virtual Address for IMMU`
I think these should probably use ~GatedStallW
`flopenrc #(P.XLEN) VAdr_iWReg (clk, reset, 1'b0, SelHPTW, VAdr_iM, VAdr_iW); //Virtual Address for IMMU`

For aesthetics, would you mind changing the signal naming convention to match the Wally style? For example change 
VAdr_iF to something like IVAdrF and PA_dM to DPAdrM?

Huda-10xe:
Why is the register enable between the Fetch and Decode stage SelHPTW and not ~StallD? I don't know much about the Design of the MMU and the HPTW. I have made changes to the tracer based on the information from the book and the wave itself. SO i observed all the signals of MMU and HPTW to ensure that the data propagates successfully for coverage collection. THe reason i asserted SelHPTW and not stall signal is beacuse the book stateds that data incase of ITLB misses goes to the HPTW and then from there back to the ITLB which exists in the fetch stage of the pieline." _**Specifically, the ITLB (itlb) asserts ITLBMissF and sends PCF to the HPTW. The page table walker handles the ITLB miss, returns PTE and PageType, and pulses ITLBWriteF to write PTE into the ITLB**_"

flopenrc #(P.XLEN) VAdr_iWReg (clk, reset, 1'b0, SelHPTW, VAdr_iM, VAdr_iW); //Virtual Address for IMMU
I think these should probably use ~GatedStallW. I want to propogate the value when HPTW is active hence the SelHPTW signal.

Sure will update the signals to match Wally Style

rosethompson:
Thanks for making these changes.  I'm going to accept the PR now, but I'm going to add a comment to SelHTPW which I think might not be correct incase we find a bug in the future.

PR #1219, opened 2025-01-02
Update copyright year to include 2025

jordancarlin:


davidharrishmc:
I’m not sure we want to touch all files to update copyright dates. @rosethompson what do you think?

rosethompson:
I think it makes more sense to update them as we edit the files.

jordancarlin:
Works for me. I know some other projects also only include the copyright years in the license file which is another option to avoid needing to change it in many places.

PR #1227, opened 2025-01-07
Update buildroot configuration method, buildroot+linux+opensbi version, and device trees

jordancarlin:
- Use defconfig files for main buildroot configuration and linux configuration. These only store any settings that are different from the defaults which makes them easier to maintain across versions and much easier to read
- Use `BR2_EXTERNAL_TREE` feature of buildroot to tell it where our configuration files are. This requires a few extra files and a reorganization to match the officially recommended layout for buildroot customization files.
- Update Linux makefile to use new external tree instead of copying files and to generate the full config from the defconfig.
- Update buildroot versions
  - Update buildroot itself from 2023.5 to 2024.11 (latest release)
  - Update linux kernel from 6.6 to 6.12.8 (latest release)
  - Update OpenSBI from 1.3 to 1.6 (latest release)
  - Use latest gcc 14.2 and binutils 2.43 for cross compilation
- Temporarily remove the `uemacs` package from buildroot because it is not compatible with gcc 14. Issue opened (https://gitlab.com/buildroot.org/buildroot/-/issues/81) and does not seem like an essential package to our buildroot image so probably fine to remove for now.
- Update linux testvector generation script to launch qemu using `fw_jump.bin` instead of `fw_jump.elf`. The `fw_jump.elf` file does not work with QEMU with newer versions of OpenSBI. See https://github.com/riscv-software-src/opensbi/issues/372
- Update device trees to include the full list of accepted extensions that wally supports. This reduces the instruction count for a linux boot and means it tests a larger portion of the code base since a wider variety of instructions are being executed. The biggest ones that I noticed it using now were bit manipulation and CMOs.


davidharrishmc:
Do you want to remove zicbop since we implement it as not?

davidharrishmc:
@rosethompson this looks good to me, but could you review as well?

davidharrishmc:
Also, is this the time to change the VCU-118 SDC in the device tree?

jordancarlin:
@rosethompson It doesn't necessarily need to happen before we merge, but it would be good to test the new version on an FPGA when you're able to.

jordancarlin:
> This looks good to me. I'm not sure it matters to include the cbop size in the device tree since it's not implemented.

Good point. Forgot to remove that when removing cbop. Gone now.

PR #1229, opened 2025-01-13
Fix buildroot when launched from install script

jordancarlin:
Generating buildroot worked when the makefile was launched directly but not when the install script called it. The issue was that `$WALLY` was not set.


jordancarlin:
Hmm. Seems like the updated version of buildroot doesn't like being built as root, which happens when running the install script. The problem seems to be specific to building tar.

davidharrishmc:
Don't love a configuration option called
FORCE_UNSAFE_CONFIGURE

Are there any alternatives?

jordancarlin:
Not a fan of it either. It's what the error message mentioned, so I'm testing it as a starting point to see if it actually fixes the problem. Will see how it goes in the CI and then investigate other options. There was nothing immediately apparent though.

jordancarlin:
Seems like with the new version of OpenSBI we need 1600000 instructions get to the OpenSBI banner instead of the previous 1400000. Updated regression-wally accordingly.

jordancarlin:
Looks like this did work to successfully install buildroot on all distros. I don't love needing to pass that configuration flag, but unfortunately I can't really think of a better option.

jordancarlin:
@davidharrishmc I still don't love the `FORCE_UNSAFE_CONFIGURE`, but I don't see another way to allow it to compile as root. While compiling as root generally isn't great, it is what we have been doing since even before I started overhauling the installation process and I don't see a way around it without making the script much more annoying to run.

PR #1244, opened 2025-01-23
Overhaul Linux build Makefile and add qemuBoot script

### LINKED ISSUE #1237: Buildroot boot fails on QEMU
The new buildroot configuration boots successfully in lockstep on Wally, but fails in QEMU. It gets through the OpenSBI boot and all associated messages, but hangs before any messages from the kernel start. It seems to be related to the Svpbmt extension, removing it from the device tree results in a successful boot.

Need to investigate if there is a way to enable this extension in QEMU or if it is just not supported. Would also be good to disable hypervisor and sdtrig extensions in QEMU to make it more closely match Wally.

jordancarlin:
- Enable missing `Svade` extension in device trees for virt and all FPGA configurations.
- Increase short buildroot instruction count to accommodate newer kernel
- Add `linux/qemuBoot.sh` script to boot Buildroot Linux in QEMU with all of the necessary flags to load the image binaries and configure QEMU to (mostly) match Wally. Main change from old flow is adding the `--cpu` flag to configure the enabled extensions.
- Remove `genInitMem.sh` script and directly integrate into Makefile. This script was always executed using `make dumptvs` anyway and the script duplicated a lot of the checks that were already performed in the Makefile.
- Restructure and simplify Linux Makefile for increased readability. Also fix some of the prerequisites.

Fixes #1237


davidharrishmc:
Looks plausible; I'll defer to Rose on these.

jordancarlin:
Just finished a successful lockstep boot, so this should be good to go once you have a chance to look at it @rosethompson.

PR #1245, opened 2025-01-23
Python updates

### LINKED ISSUE #1176: Investigate a way to test python scripts for compatibility with all supported versions in install script
Depending on the distro being used, we install different versions of tools. For example, we install versions of python between 3.9 and 3.12. Investigate automated way of ensuring our scripts our compatible with all of these versions. This has already caused some issues for @rosethompson with python 3.12, and when working on the updated version of wsim I ran into the opposite issue and attempted to use a feature not present in python 3.9.

Investigating adding some kind of python linting as a CI check. Potentially using Ruff.

jordancarlin:
- Refactor most of the python files to use more standard python conventions and conform to the ruff linter
- WIP: enable ruff lint in CI to ensure all python code is compatible with all needed python versions
  - Also make it possible to run this locally so errors can be discovered before pushing to GitHub

closes #1176


davidharrishmc:
It's a lot of changes.  They look reasonable, but do we have a way of knowing they don't introduce new bugs, especially in scripts that aren't part of a regression process?

jordancarlin:
That’s why it’s still a draft. I’m still changing some of the files, but I’ll be running each of the scripts with all of its potential flags/options before we merge to make sure noting broke.

jordancarlin:
@davidharrishmc This should be good to go now. I've finished updating the remaining python scripts and the python lint is now passing when a PR is opened. I'll add local python linting later as a follow up PR. I've tested all of the scripts that I modified and they all either succeed or give the same error messages as before (in the case of some of the synth scripts).

davidharrishmc:
The changes in the riscof code are in files that were not originally ours.  Is it necessary for us to change these?  If so, should we be trying to push any of these changes upstream?

jordancarlin:
> The changes in the riscof code are in files that were not originally ours. Is it necessary for us to change these? If so, should we be trying to push any of these changes upstream?

It looks like our versions have already diverged a decent bit and there will need to be a bit of an overhaul to these files (both upstream and in our repo) in the near future to properly support PMP and to work with the new Sail config system. I'd say for now let's merge these and I can work on upstreaming most of our changes and then getting us back to a version that is more closely in line with the upstream version. It'd be nice if we can eventually get to the point of directly using the upstream versions so we get all of the latest changes (we already have them in the riscv-arch-test submodule).

PR #1247, opened 2025-01-24
Add Makefile in testbench to allow the memfile to generate the correc…

stineje:
…t version based on the elf output in objdump.  Previously this was done via a findstring which works with riscv-arch-tests but doesn't allow individual programs/elf to be used unless the program is called xxx_rv32.

Tested with risv-arch-test as well as 32/64 bit programs in examples.  All generate correct response to allow simulation to work correctly.


jordancarlin:
Might be worth updating all of the other Makefiles that have a similar rule as well. This seems like a much more reliable method of finding the bit width.

PR #1249, opened 2025-01-26
Cleanup

jordancarlin:
- Remove empty `checksignature.sv` file (it just had a header and no actual content)
- Rename `FunctionName` module to `functionName` to match the module name to the filename and to be consistent with all of our other module naming conventions.


davidharrishmc:
Do you want to commit debug=1

jordancarlin:
Oops thought I undid that. Fixed.

PR #1253, opened 2025-01-29
Updating Wally Tracer for MMU signal Propagation

Huda-10xe:
SelHPTW was affecting both StallM and StallW. Initially we hadn't taken into account its affect on StallM but that has been rectified now.


davidharrishmc:
Good!  Looks like Rose suggested several other signals need stall gating too.  Can you two explore all these at once and fix them together?  Otherwise, we might be getting spurious coverage and not know it.

Huda-10xe:
@rosethompson would you like to have a meeting to go over these once just to confirm if they're working properly or do you reckon that if they are collecting coverage for all sv32 tests they would be working alright?

rosethompson:
I suspect if they are collecting the coverage you are expecting they are probably working correctly. But I think it would be worth going over a review to make sure it is correct. Are there other changes needed to the tracer?  If so maybe we group these into one meeting?

Issue #1251, opened 2025-01-29
Unused coverage files?


jordancarlin:
From what I can tell, the functional coverage files at `testbench/coverage` are not used and have likely been superseded by `cvw-arch-verif`. Looks like they were attempts at coverage for fences and pmps. Any reason not to remove these?

jordancarlin:
Closed with #1254

PR #1254, opened 2025-01-29
Removed unused testbench/coverage directory

davidharrishmc:
Fixes Issue #1251


jordancarlin (changes requested):
Overall looks good, but I think there is an issue with `NUM_REGS`. From what I can tell, an rv32ef core will still have 32 fp registers, so everything relating to floating point should not use `NUM_REGS`. It looks like `frf` itself is set up correctly using 32, but all the other related variables are using `NUM_REGS`.

davidharrishmc:
Great catch!
Pushed a fix.

jordancarlin (approved):

PR #1256, opened 2025-01-30
Refactor functional coverage

jordancarlin:
- Updates to testbench, simulation scripts, and config files for compatibility with the overhauled functional coverage in cvw-arch-verif.
- Functional coverage and lockstep are now two discrete flags for wsim.


jordancarlin:
This should finally be ready to go! The one thing we still need to decide is if we want `regression-wally --fcov` to just run functional coverage or to run functional coverage and lockstep simultaneously. `wsim` will definitely have separate flags, but we could go either way for `regression-wally`.

PR #1258, opened 2025-01-30
Sail cmake

jordancarlin:
Sail has updated to using cmake for its build process. The simulator was also renamed to indicate which flen it was compiled with, so a "d" needs to be added to the simulator name. Also install newer CMake on old Debian and Ubuntu distros.


jordancarlin:
CI failing until #1229 is merged

jordancarlin:
Rebased on top of #1229. Should be good once CI passes (let's wait for the though).

jordancarlin:
Seems there were a couple of other things that have broken the installation. Will work on a fix.

jordancarlin:
Looks like that generally fixed things, but now it’s failing on Debian 11 because of an incompatible version of cmake (which is exactly what we set up the CI to try and catch). I’ll take a look at whether we need to make changes upstream or just install a newer version of cmake manually.

davidharrishmc:
Beautiful!

jordancarlin:
Turns out sail does indeed rely on some of the newer features of CMake. I updated the script to directly install the newer version for Debian 11 and Ubuntu 20.04. Hoping we'll get successful CI now.

jordancarlin:
Once we merge this we’ll need to update Sail on all our servers or make will start failing because of the rename.

PR #1260, opened 2025-02-02
Always run installation CI in clean docker image

### LINKED ISSUE #1259: Ubuntu CI has newer packages
The installation CI is using the default GitHub Actions runners for Ubuntu installation tests. While working on #1258, I discovered that these images come with newer versions of some packages that have been manually installed and are not normally available in the default Ubuntu package repos.

We should update the Ubuntu installation CI runs to directly use a clean Ubuntu docker image to avoid this. It led me to believe that the new version of sail worked, but that was actually only because GitHub preinstalls CMake 3.31 (3.16 is what is actually available from apt).

jordancarlin:
Fixes #1259

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -37,14 +37,14 @@ jobs:
         include:
           # Ubuntu Installations
           - name: ubuntu-20.04
-            container: ubuntu:20.04
+            image: ubuntu:20.04
             imageFamily: debian
             regressionFail: true
           - name: ubuntu-22.04
-            container: ubuntu:22.04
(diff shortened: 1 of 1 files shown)


jordancarlin:
Well that was unexpected. Turns out the version of git that actually comes with Ubuntu 20.04 is too old to clone riscv-gnu-toolchain.

davidharrishmc:
Time to drop 20.04?

jordancarlin:
@davidharrishmc Let's merge this for now to switch over how the CI system works and then I will have a follow up PR that deprecates Ubuntu 20.04.

PR #1262, opened 2025-02-04
Deprecate support for Ubuntu 20.04

jordancarlin:


jordancarlin:
@davidharrishmc Heard back from someone who had a similar issue with Ubuntu 20.04, and turns out there is a flag you can pass to git configure that makes it work (telling it to use the newer protocol version). Do we still want to proceed with removing support for Ubuntu 20.04 or do we add this flag and keep it alive for now?

jordancarlin:
Ubuntu support fixed in #1274, so closing this for now. Will revisit in May when Ubuntu 20.04 reaches EOL.

PR #1265, opened 2025-02-07
Initial FMA commit

davidharrishmc:


examples/exercises/fma16/lint-fma
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# check for warnings in Verilog code
+# The verilator lint tool is faster and better than Questa so it is best to run this first.
+export PATH=$PATH:/usr/local/bin/
```
jordancarlin: Not sure it makes sense to include this in the lint script. Verilator can be installed to many different locations (on our servers it’s at `/opt/riscv/bin`.

davidharrishmc: Removed

PR #1268, opened 2025-02-08
E154 Lab 2 fma16_testgen and odd solutions

davidharrishmc:


examples/exercises/fma16/Makefile
```diff
@@ -0,0 +1,16 @@
...
+CC     = gcc
+CFLAGS = -O3 -Wno-format-overflow
+IFLAGS = -I$(WALLY)/addins/berkeley-softfloat-3/source/include/
+LIBS   = $(WALLY)/addins/berkeley-softfloat-3/build/Linux-x86_64-GCC/softfloat.a -lm -lquadmath
+SRCS   = $(wildcard *.c)
+PROGS = $(patsubst %.c,%,$(SRCS))
+
+all:	$(PROGS)
+
+%: %.c
```
jordancarlin, line 12: This should probably have a dependency on the softfloat library to ensure it is compiled before attempting to compile this.

.gitignore
```diff
@@ -168,3 +168,8 @@ config/deriv
 sim/slack-notifier/slack-webhook-url.txt
 docs/docker/buildroot-config-src
 docs/docker/testvector-generation
+examples/C/fmul
```
jordancarlin: Move these to the examples section?

jordancarlin:
Looks like the generated `fma16_testgen` executable expects a `work` directory to exist and fails without it. Would be good to have it create this directory.

examples/exercises/fma16/Makefile
davidharrishmc: Other files in fp/fpcalc don't have this dependency.  How about merge this PR, and then add it everywhere if you want it.

jordancarlin: Works for me

jordancarlin (approved):

PR #1279, opened 2025-02-16
Simplify coverage selection

jordancarlin:
Updates to accompany https://github.com/openhwgroup/cvw-arch-verif/pull/450


davidharrishmc:
Not sure I understand this one. What if I want zfh without d or zcb without m?

davidharrishmc:
Oh I see

jordancarlin:
Take a look at the PR in the other repo. It only enables the feature if both extensions are enabled.

PR #1292, opened 2025-02-22
Added various covergroups to fcov

coreyqh:


config/rv32gc/coverage.svh
```diff
@@ -53,7 +53,17 @@
...
 `define EXCEPTIONSZC_COVERAGE
+`define EXCEPTIONSZAAMO_COVERAGE
+`define EXCEPTIONSZALRSC_COVERAGE
+`define EXCEPTIONSZICBOS_COVERAGE
+`define EXCEPTIONSZICBOU_COVERAGE
 `define ZICNTRU_COVERAGE
 `define ZICNTRS_COVERAGE
 `define ZICNTRM_COVERAGE
+`define INTERRUPTSU_COVERAGE
+// `define INTERRUPTSM_COVERAGE
```
jordancarlin: With the new config system, I think you should be good to uncomment these and it shouldn't cause issues. They shouldn't do anything until they are merged in cvw-arch-verif.

PR #1299, opened 2025-02-25
Non-zero DAU bits in non-leaf PTE cause page faults.

rosethompson:
Makes Wally consistent with Breker's interpretation of the privilege spec. This portion of the spec clarifies previous vagueness. 

> If pte.v = 0, or if pte.r = 0 and pte.w = 1, or if any bits or encodings that are reserved for future standard use are set within pte, stop and raise a page-fault exception corresponding to the original access type.

### DIFF
--- a/src/mmu/hptw.sv
+++ b/src/mmu/hptw.sv
@@ -104,18 +104,21 @@ module hptw import cvw::*;  #(parameter cvw_t P) (
   logic                     HPTWAccessFaultDelay;
   logic                     TakeHPTWFault;
   logic                     PBMTFaultM;
+  logic                     DAUFaultM;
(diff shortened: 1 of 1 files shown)


davidharrishmc:
I have a PR ready too.  It did not pass coverage64gc.  I’ll compare when I’m not in class.

rosethompson:
I was struggling to get the regression tests built, but Jordan just pointed me in the correct direction so I should be able to get it running soon.

rosethompson:
Looks like the problem is the ebu.S test doesn't expect to generate page faults on non-zero DAU bits.

davidharrishmc:
I came to essentially the same code.  Merging this one.

PR #1312, opened 2025-03-11
Code coverage fixes

jordancarlin:
- Fix code coverage collection in regression-wally
- Fix wsim failure edge case with elf files without an extension in lockstep
- Generate code coverage report for hazard unit

### DIFF
--- a/bin/regression-wally
+++ b/bin/regression-wally
@@ -514,7 +514,7 @@ def main(args):
 
     # Coverage report
     if args.ccov:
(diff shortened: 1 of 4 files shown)


davidharrishmc:
What are the implications of removing the volatile from the H counters?

jordancarlin:
That was not supposed to be in this PR. Will fix soon. Thanks for catching that.

jordancarlin:
> What are the implications of removing the volatile from the H counters?

These lines are duplicated at 850. Removing them was actually intentional.

Issue #1315, opened 2025-03-12
rv32imc Lockstep Mismatches from when S was enabled


jordancarlin:
- Wally bugs
    - [ ] Illegal instructions don’t trap
- Imperas configuration issues or bugs
    - [ ] mstatus.SUM not read-only zero when SATP read only zero
      - The SUM bit should be read-only 0 when SATP is read-only 0. I've set `--override cpu/Sv_modes=1`, which should restrict virtual memory to bare mode only and therefore disable this bit.
- Other bugs
    - [ ]  `csrrs x5,satp,x0` is causing an illegal instruction exception on Wally and not on Imperas. Further investigation still needed to determine correct behavior.

jordancarlin:
These bugs are no longer present now that S is disabled. They are worth revisiting at some point, but no longer high priority.

davidharrishmc:
Fixed the SUM read-only zero, and some other read-only zeros while I'm at it.  The others seem like mismatches with the Imperas configuration, which don't matter because we don't actually have an Imperas config now for 32IMCS.  Can we close this?

jordancarlin:
Seems reasonable to close since we no longer support this configuration. Not sure if the lack of trap issue was sorted out, but that can be addressed if a configuration like this is ever added.

PR #1316, opened 2025-03-13
Add rv32IMC Imperas Configuration for Lockstep

jordancarlin:
Currently fails several tests in lockstep due to problems explained in #1315, but this should be good to merge as is.

Also adds coverage config file for rv32imc


davidharrishmc:
There is no reason for a microcontroller to have S mode without VM.  Table A1. calls rv32imc MU, but table A.3 has a 1 in S mode for rv32imc.  I’ll change table 3.  In coverage.svh, please drop the S coverage.  In config.vh, please remove S.
Zifencei is a nop in systems without caches, and there’s no way to synchronize an IROM with a DTIM.  Let’s change ZifenceI to 0 in config.vh for rv32imc and strike it from coverage.svh.
In imperas.ic
Disable supervisor mode.
Why do we set wfi_is_nop in all our config files?  We really do wait.

jordancarlin:
Disabled S and Zifencei for rv32imc config of Wally. The `wally32periph` suite now fails. We need to either drop it from the rv32imc tests or modify it to work without supervisor mode. @davidharrishmc opinions on this?

davidharrishmc:
I don’t recall why 32periph fails without supervisor. Maybe plic is making
seip?  Perhaps there is one test we could pull out?  It would be good to
keep most of period on 32imc to keep it working as a microcontroller.

jordancarlin:
Most of the periph tests are already pulled out. gpio, clint, uart, plic, plic-s, and spi are all part of wally32priv instead of wally32periph because they apparently don't work on rv32imc because of the lack of pmp. The only test in the wally32periph suite is `WALLY-periph-01.S` itself.

jordancarlin:
Looks like it is a similar issue to https://github.com/openhwgroup/cvw-arch-verif/issues/531. `WALLY-periph-01.S` does a `csrrc` to `mideleg`, which doesn't exist without S-mode. The trap handler then proceeds normally but returns to the same instruction that trapped, which of course immediately traps again.

jordancarlin:
wally32periph test removed for now (see #1321). There are still issues with some of the functional coverage tests, but those likely require changes in cvw-arch-verif. This should be ready to merge for an initial rv32imc lockstep implementation.

PR #1318, opened 2025-03-13
100% fctrl and Hazard code coverage

coreyqh:
Minor change needed to RTL, since the condition on legal `Fmt` meant two of the conditions in the line were redundant. Still passes regression-wally --fcov as it should be functionally equivalent.


davidharrishmc:
* The old and new fctrl code are not equivalent.  The old code forbids fmv.x.q, while the new code would allow it.  The multiple negation in the new code is also hard to read.

* Is the new fpu.S code required when running with —fcov? fcov should try all the unsupported dynamic rounding modes, so this fpu.S code shouldn’t be necessary. In general, I’d rather improve functional coverage than tests/coverage if possible.

davidharrishmc:
For excluding coverage of the Q instructions, it's better to put it in the coverage exclusion file for just the Q scenario rather than to // coverage off the entire line.  See examples of how to do this:

https://github.com/openhwgroup/cvw/blob/main/sim/questa/coverage-exclusions-rv64gc.do

src/hazard/hazard.sv
```diff
@@ -101,13 +101,19 @@ module hazard (
...
+  // StallD = StallF so LatestUnstalledD is necessarily 0
   assign LatestUnstalledD = ~StallD & StallF;
+  // coverage on 
   assign LatestUnstalledE = ~StallE & StallD;
   assign LatestUnstalledM = ~StallM & StallE;
   assign LatestUnstalledW = ~StallW & StallM;
   
   // Each stage flushes if the previous stage is the last one stalled (for cause) or the system has reason to flush
+  // coverage off
+  // LatestUnstalledD is always 0
```
jordancarlin: Should we just be excluding LatestUnstalledD so that we still cover FlushD being both 0 and 1 from FlushDCause?

coreyqh: Sounds reasonable. Should I then also change the line that assigns StallF from StallD and StallFCause, which is presently excluded?

jordancarlin: I think that signal is always 0, so less of a strong opinion on that one. It wouldn't hurt I guess though.

davidharrishmc: Let's use coverage tags with exclusions instead of coverage on/off anywhere feasible.  That would include FlushD.

PR #1337, opened 2025-04-01
Rv64 Config Changes

C0d3ing:
Added the remaining Exceptions tests to Rv64gc's coverage config file


davidharrishmc:
Why are these only in RV64?

PR #1345, opened 2025-04-08
add Python script used to analyze an ELF file to see instructions from ELF file

stineje:
Wanted to create a script for use in my class to do a histogram on an ELF file.  Seem to work well - ran on CoreMark:

```
Instruction usage in work/coremark.bare.riscv:

addi       3,097
lw         973
beq        867
sw         714
add        538
jal        499
lbu        329
```

### DIFF
--- a/bin/analyze_riscv_elf.py
+++ b/bin/analyze_riscv_elf.py
@@ -31,10 +31,10 @@
 ## and limitations under the License.
 ################################################################################################
 
-from collections import Counter
+import re
 import subprocess
 import sys
-import re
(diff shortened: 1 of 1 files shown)


davidharrishmc:
Looks good.  Can you make it pass the Lint above?

jordancarlin:
Good reminder that I should add the pre-commit hook for ruff to our flow so that it will automatically fix these simple things (or at least warn you that there are issues) when you commit.

rosethompson:
Can you also add the SPDX header?

PR #1349, opened 2025-04-09
Enable pre-commit hooks

jordancarlin:
Add pre-commit hooks to ensure all lint rules pass before committing. This avoids repeatedly pushing to GitHub to try resolving a lint error.

The setup script is updated to automatically enable the pre-commit hooks whenever someone clones the repo and the installation files are updated to install `pre-commit` when running the main toolchain install.

Migrates the existing ruff lint to run in pre-commit hooks and adds several other basic checks (no large files, no merge conflict indicator marks, parseable yaml/json files, etc.).

There are several other pre-commit hooks that we should enable, but they touch a decent number of files so I'll do them as separate PRs to make it easier to review.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   lint:
-    name: Python ${{matrix.version}} lint
+    name: Lint (Python ${{matrix.version}})
     strategy:
       matrix:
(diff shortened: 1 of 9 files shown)


davidharrishmc:
This scares me a little bit.  If you ever graduate and something breaks, I'll never figure out how to fix it.

PR #1351, opened 2025-04-09
Full csru code coverage

coreyqh:
Functional coverage suite attempts writes to fcsr, frm, fflags with mstatus.FS = 0 but this was not covered. Further inspection reveals that this attempt traps and flushes the M stage and therefore cannot be covered. 
(FS != 0) -> IllegalCSRUAccessM -> IllegalCSRAccessM -> IllegalInstrFaultM -> ExceptionM -> trapM -> FlushM


davidharrishmc:
This looks good.  Can you add a line in csru.sv below 53 indicating that the STATUS_FS check is redundant because if it is 0, the instruction will trap and CSRUWriteM will be low anyway, but that the check is still here out of caution.

PR #1355, opened 2025-04-11
Fixed mstatus bits that should be read-only 0 in certain configurations

davidharrishmc:
Fixes part of https://github.com/openhwgroup/cvw/issues/1315

### DIFF
--- a/src/privileged/privdec.sv
+++ b/src/privileged/privdec.sv
@@ -72,7 +72,7 @@ module privdec import cvw::*;  #(parameter cvw_t P) (
   assign ebreakM =    PrivilegedM & (InstrM[31:20] == 12'b000000000001) & rs1zeroM;
   assign wfiM =       PrivilegedM & (InstrM[31:20] == 12'b000100000101) & rs1zeroM;
   assign sfencevmaM = PrivilegedM & (InstrM[31:25] ==  7'b0001001 | invalM) & 
-                      (PrivilegeModeW == P.M_MODE | (PrivilegeModeW == P.S_MODE & ~STATUS_TVM)); 
+                      (PrivilegeModeW == P.M_MODE | (PrivilegeModeW == P.S_MODE & ~STATUS_TVM)) & P.VIRTMEM_SUPPORTED; 
 
   ///////////////////////////////////////////
   // WFI timeout Privileged Spec 3.1.6.5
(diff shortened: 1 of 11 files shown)


davidharrishmc:
Also fixes regression WALLY-trap tests that stopped liking an illegal instruction of 00000000.  Unsure why this compressed instruction started causing mismatches or why it even worked in the past, but now FFFFFFFF works.

src/privileged/privdec.sv
```diff
@@ -72,7 +72,7 @@ module privdec import cvw::*;  #(parameter cvw_t P) (
   assign ebreakM =    PrivilegedM & (InstrM[31:20] == 12'b000000000001) & rs1zeroM;
   assign wfiM =       PrivilegedM & (InstrM[31:20] == 12'b000100000101) & rs1zeroM;
   assign sfencevmaM = PrivilegedM & (InstrM[31:25] ==  7'b0001001 | invalM) & 
-                      (PrivilegeModeW == P.M_MODE | (PrivilegeModeW == P.S_MODE & ~STATUS_TVM)); 
+                      (PrivilegeModeW == P.M_MODE | (PrivilegeModeW == P.S_MODE & ~STATUS_TVM)) & P.VIRTMEM_SUPPORTED;
```
jordancarlin, line 75: The spec says 
> For implementations that make satp.MODE read-only zero (always Bare), attempts to execute an SFENCE.VMA instruction **might** raise an illegal-instruction exception.

So seems like either way is a valid implementation choice. Just want to confirm this is a deliberate change to how Wally chooses to handle this.

testbench/common/riscvassertions.sv
```diff
@@ -62,6 +62,7 @@ module riscvassertions import cvw::*; #(parameter cvw_t P);
     assert ((P.ZICBOZ_SUPPORTED == 0) | (P.DTIM_SUPPORTED == 0)) else $fatal(1, "ZICBOZ incompatible with DTIM");
     assert ((P.SVPBMT_SUPPORTED == 0) | (P.VIRTMEM_SUPPORTED == 1 & P.XLEN==64)) else $fatal(1, "SVPBMT requires VIRTMEM_SUPPORTED and RV64");
     assert ((P.SVNAPOT_SUPPORTED == 0) | (P.VIRTMEM_SUPPORTED == 1 & P.XLEN==64)) else $fatal(1, "SVNAPOT requires VIRTMEM_SUPPORTED and RV64");
+    assert ((P.SVINVAL_SUPPORTED == 0) | (P.VIRTMEM_SUPPORTED == 1)) else $fatal(1, "SVINVAL requires VIRTMEM_SUPPORTED and RV64");
```
jordancarlin: There is no RV64 constraint here.
```suggestion
    assert ((P.SVINVAL_SUPPORTED == 0) | (P.VIRTMEM_SUPPORTED == 1)) else $fatal(1, "SVINVAL requires VIRTMEM_SUPPORTED");
```

src/privileged/privdec.sv
davidharrishmc: Yes, moving toward strictness.

jordancarlin: Great.

PR #1365, opened 2025-04-17
Full priv and cacheways code coverage

coreyqh:
Two exclusions for impossible expression rows that are guaranteed to trap, and a single trivial test for a successful user mode CSR write.


davidharrishmc:
ifu.S
.hword 0x8002 shouldn’t be necessary when #559 is merged
priv.S
Writing user CSR in U-mode should be tested by ZicsrU.S.  However, it isn’t because ZicsrU didn’t turn on floating-point and the accesses trapped.  Fixed with #559; should be able to drop this.
dcache2.S
lr and sc are unusual.  Confirming they are necessary here.  If so, ignore this comment.

PR #1366, opened 2025-04-17
Cleaned up HPTW reset

davidharrishmc:


jordancarlin:
Oops. Not sure what I hit with Copilot there.

PR #1367, opened 2025-04-17
Branch predictor benchmarking/performance validation updates

jordancarlin:
- Add `regression-wally --branch` to run embench on all of the branch predictor configurations
  - Enable hpm logging for them
- Update list of branch configs to use in sweep
- Fix paths to be simulator agnostic

`parseHPMC.py` still giving strange results with some of the generated graphs. Need to continue investigating, but this should be good to merge for now.

@rosethompson Can you look over the updated list of branch configs that are being tested for each sweep?


rosethompson:
Can we make a backup of branch-list-global-local.txt, I don't want to lose the contents.

jordancarlin:
> Can we make a backup of branch-list-global-local.txt, I don't want to lose the contents.

That is a new file that I created in this PR. It is a copy of all of the global/local entries from the old `branch-list.txt` file since those were removed.

rosethompson:
I'm sorry I meant to write branch-list.txt. I understand now.  This is ok for now.

PR #1370, opened 2025-04-18
Add +acc flag to Questa when dumping vcd and fix no VM derived configs

### LINKED ISSUE #1178: Questa VCD dump not generating correctly
Questa VCD dump is significantly smaller than the dumps from Verilator and VCS:

>Verilator: 4 GB
Questa: 9 MB
VCS: 6.3 GB

Likely that it is not logging at the right level.

jordancarlin:
Seems to fix #1178. The vcd dump from Questa is now several GB, as expected.


jordancarlin:
Also updated derivlist to disable Svinval on configs without virtual memory. This was causing nightly regression failures after the new assertion was added recently.

PR #1378, opened 2025-04-22
fma16 project synthesis and test vectors

davidharrishmc:


jordancarlin:
Looks like the README needs to be updated. It seems to be referring to synthesizing Wally, not the fma.

examples/exercises/fma16/synthDC/README.md
```diff
@@ -0,0 +1,25 @@
...
+## Example Usage
+```bash
+make synth FREQ=500 
+```
+
+## Environment Variables
+
+- `FREQ`
+  - Frequency in MHz.  Default is 500
+  - The target standard cell library.  The default is sky130.
```
jordancarlin, line 16:
```suggestion
- `TECH`
    - The target standard cell library.  The default is sky130.
```

examples/exercises/fma16/synthDC/README.md
```diff
@@ -0,0 +1,26 @@
...
++- `TECH`
```
jordancarlin: With the plus it appears as part of the previous line instead of as a new bullet.

```suggestion
- `TECH`
```

PR #1390, opened 2025-04-26
Performance Benchmarking Script - Using grep and other internal bash commands within regression-wally

SadhviNarayanan:
Performance Benchmarking Script - Using grep and other internal bash commands to run validation within regression-wally
- Performance metrics for the coremark and embench are validated against expected values using bash commands within regression-wally
- This approach makes it easier to maintain, however it is not as modular and scalable as tests need to become more robust


bin/regression-wally
```diff
@@ -394,6 +398,8 @@ def parse_args():
     parser.add_argument("--branch", help="Run branch predictor accuracy tests", action="store_true")
     parser.add_argument("--breker", help="Run Breker tests", action="store_true") # Requires a license for the breker tool. See tests/breker/README.md for details
     parser.add_argument("--dryrun", help="Print commands invoked to console without running regression", action="store_true")
+
+    parser.add_argument("--performance", help="Check for performance changes or discrepencies in embench and coremark", action="store_true")
```
jordancarlin:
```suggestion
    parser.add_argument("--performance", help="Check for performance changes or discrepancies in embench and coremark", action="store_true")
```

bin/regression-wally
```diff
@@ -414,15 +420,15 @@ def process_args(args):
...
+    elif args.buildroot or args.performance: # TODO: fix timing on this bc performance shodl also be limit on nightly
         TIMEOUT_DUR = 60*3600 # 2.5 days
     elif args.testfloat:
         sims = [testfloatsim]
         TIMEOUT_DUR = 30*60 # seconds
     elif args.branch:
         TIMEOUT_DUR = 120*60 # seconds
     elif args.nightly:
-        TIMEOUT_DUR = 30*60 # seconds
+        TIMEOUT_DUR = 5*3600 # seconds # NOTE: changed this to 5 hours for nightly regression
```
jordancarlin: Can probably remove this note.

bin/regression-wally
```diff
@@ -414,15 +420,15 @@ def process_args(args):
         TIMEOUT_DUR = 30*60
         shutil.rmtree(f"{regressionDir}/questa/fcov_ucdb", ignore_errors=True)
         os.makedirs(f"{regressionDir}/questa/fcov_ucdb", exist_ok=True)
-    elif args.buildroot:
+    elif args.buildroot or args.performance: # TODO: fix timing on this bc performance shodl also be limit on nightly
```
jordancarlin: Should the `or args.performance` be with `args.nightly` instead? I assume the performance tests don't need 2.5 days?

bin/regression-wally
```diff
@@ -504,6 +510,66 @@ def selectTests(args, sims, coverStr):
                         grepstr="All Tests completed with          0 errors",
                         grepfile = sim_log)
                 configs.append(tc)
+
+    if (args.performance or args.nightly):
+        # RUNNING THE EMBENCH TEST
+        json_dir_1 = f"{WALLY}/benchmarks/embench/wallySizeOpt_size.json"
```
jordancarlin: These aren't directories. Should this be named `json_1` instead? Or even `wallySizeOpt_size_json`.

bin/regression-wally
```diff
@@ -504,6 +510,66 @@ def selectTests(args, sims, coverStr):
...
+        f"|| (echo 'EMBENCH_TEST_FAILED: Values did not match expected metrics' >> $WALLY/benchmarks/embench/run.log && exit 1)"
+    ),
+        
+            grepstr="EMBENCH_TEST_PASSED", # not sure what keyword to put here --> could also put success which comes from the simulation
+            grepfile=os.path.expandvars("$WALLY/benchmarks/embench/run.log")
+        )
+        configs.append(embench_test)
+
+        # RUNNING THE COREMARK TEST
+        validation_log = f"{WALLY}/bin/logs/validation.log"
```
jordancarlin: Let's keep logs out of the `bin` directory. It should only have scripts in it. How about using either the `embench` directory or the sim log directory corresponding to whatever simulator is being used?

bin/regression-wally
```diff
@@ -504,6 +510,66 @@ def selectTests(args, sims, coverStr):
...
+        os.makedirs(os.path.dirname(validation_log), exist_ok=True)
+
+        coremark_sweep_test_dir = f"{WALLY}/benchmarks/coremark/coremark_sweep.py"
+
+        actual_coremark_values_csv = f"{WALLY}/benchmarks/coremark/coremark_results.csv"
+        expected_coremark_values_csv = f"{WALLY}/benchmarks/coremark/expected_coremark_results.csv"
+
+        performance_test = TestCase(
+            name="validate_performance",
```
jordancarlin: Should this be named `coremark`? I think it is just testing coremark and not all of the performance validation.

bin/regression-wally
```diff
@@ -504,6 +510,66 @@ def selectTests(args, sims, coverStr):
...
+            variant="performance check",
+            cmd=(
+                "set -e && "
+                f"touch {validation_log} && "
+                f"python3 {coremark_sweep_test_dir} && "
+                f"if diff -q {actual_coremark_values_csv} {expected_coremark_values_csv}; then "
```
jordancarlin: I wonder if we could simplify this a bunch by adding another entry to the `TestCase` tuple. Something like `checkcmd`. It could default to grepping but if set would perform that command instead. That way you could have that be the diff command and not need a kind of hacky workaround of echoing things into a log. This is a great approach if we want to keep the current structure though!

jordancarlin (commented):
Looking great so far! Here are some initial comments.

bin/regression-wally
SadhviNarayanan: yes! i think I meant to change this, but because I missed this TODO I did not - it should be a part of the nightly

bin/regression-wally
SadhviNarayanan: found a way to implement this

bin/regression-wally
SadhviNarayanan: renamed

PR #1391, opened 2025-04-28
Updated the script to run all verification within regression-wally

SadhviNarayanan:
Updated the script to run all verification within regression-wally
- Implemented more concise logic
- Fixed log file paths
- Cleaned up code


davidharrishmc:
expected_results.json doesn’t belong in bin.  How about in benchmarks?  Or separate files in benchmarks/coremark and benchmarks/embench?

There are expected numbers built into the regression-wally script.  Can these numbers be read from the expected_results file rather than hardwired in a script?

5 hours is a long timeout for nightly, and may leave other bad jobs hanging for a long time.  Would it make sense to separate the benchmarking into multiple smaller jobs that can be run independently?  Or only benchmark one config during nightly?

There’s a moderate amount of new stuff in regression-wally. In terms of maintainability, the less, the better.  Is this the simplest necessary?

bin/regression-wally
```diff
@@ -502,8 +533,67 @@ def selectTests(args, sims, coverStr):
...
+        wallySpeedOpt_size_json = f"{WALLY}/benchmarks/embench/wallySpeedOpt_size.json"
+        wallySpeedOpt_speed_json = f"{WALLY}/benchmarks/embench/wallySpeedOpt_speed.json"
+
+        embench_test = TestCase(
+            name="embench",
+            variant="rv32gc",  # is this the correct variant here? or rv32imac_zicsr 
+            cmd=(
+        f"touch $WALLY/benchmarks/embench/run.log && ("
+        f"cd $WALLY/benchmarks/embench && make run | tee run.log && "
+        f"grep -q 'size geometric mean.*1.04' {wallySizeOpt_size_json} && "
```
jordancarlin: Any reason not to just have this do a diff like Coremark so that it doesn’t need as much hardcoded stuff?

bin/regression-wally
```diff
@@ -287,7 +289,10 @@ TestCase = namedtuple("TestCase", ['name', 'variant', 'cmd', 'grepstr', 'grepfil
 #           grep finds that string in the logfile (is used by grep, so it may
 #           be any pattern grep accepts, see `man 1 grep` for more info).
 # grepfile:  a string containing the location of the file to be searched for output
-
+# altcommand:  the command to run, if enabled, to check the output of the CoreMark benchmark
```
jordancarlin: Altcommand isn't specific to Coremark. I think this comment needs to be updated.

bin/regression-wally
```diff
@@ -316,7 +321,9 @@ def addTests(testList, sim, coverStr, configs):
                     variant=config,
                     cmd=f"{cmdPrefix} {t} > {sim_log}",
                     grepstr=gs,
-                    grepfile = grepfile)
+                    grepfile = grepfile,
+                    altcommand=None,
```
jordancarlin: I think there is a way to have default values when you define the tuple so you don't need to add this to almost all of the instances of it.

bin/regression-wally
```diff
@@ -394,6 +422,8 @@ def parse_args():
     parser.add_argument("--branch", help="Run branch predictor accuracy tests", action="store_true")
     parser.add_argument("--breker", help="Run Breker tests", action="store_true") # Requires a license for the breker tool. See tests/breker/README.md for details
     parser.add_argument("--dryrun", help="Print commands invoked to console without running regression", action="store_true")
+
```
jordancarlin: Extra whitespace

bin/regression-wally
```diff
@@ -502,8 +534,108 @@ def selectTests(args, sims, coverStr):
...
+        # RUNNING THE COREMARK TEST
+        sim_log = f"{regressionDir}/{defaultsim}/logs/validation.log"
+        # Create the directory if it doesn't exist
+        os.makedirs(os.path.dirname(sim_log), exist_ok=True)
+
+        # Clear the file contents for a fresh start
+        with open(sim_log, 'w') as f:
+            f.write('')  # clears the file
+
+        coremark_sweep_test_dir = f"{WALLY}/benchmarks/coremark/coremark_sweep.py"
```
jordancarlin: This isn't a directory (which is what it indicates with the variable name).

bin/regression-wally
```diff
@@ -502,8 +534,108 @@ def selectTests(args, sims, coverStr):
...
+        configs.append(embench_test)
+
+
+        # RUNNING THE COREMARK TEST
+        sim_log = f"{regressionDir}/{defaultsim}/logs/validation.log"
+        # Create the directory if it doesn't exist
+        os.makedirs(os.path.dirname(sim_log), exist_ok=True)
+
+        # Clear the file contents for a fresh start
+        with open(sim_log, 'w') as f:
```
jordancarlin: regression deletes everything in the log directory when it starts so this shouldn't be necessary. Same with creating the directory above.

jordancarlin (commented):
Seems like it might be simpler to store the embench expected result files in the repo unmodified and use diff to compare them (like you're doing with Coremark already) to avoid all of the complicated json parsing logic. That might be more maintainable and avoids as much special casing.

benchmarks/embench/expected_results.json
jordancarlin, line 1: Should this file be deleted now?

bin/regression-wally
```diff
@@ -501,8 +527,88 @@ def selectTests(args, sims, coverStr):
...
+            grepfile=embench_logfile_path,
+            simlog=embench_logfile_path
+        )
+        configs.append(embench_test)
+
+
+        # RUNNING THE COREMARK TEST
+        sim_log = f"{regressionDir}/{defaultsim}/logs/validation.log"
+        # Create the directory if it doesn't exist
+        os.makedirs(os.path.dirname(sim_log), exist_ok=True)
```
jordancarlin: This directory is already created above.

bin/regression-wally
```diff
@@ -278,7 +279,7 @@ lockstepwaivers = [
 # Data Types & Functions
 ##################################
 
-TestCase = namedtuple("TestCase", ['name', 'variant', 'cmd', 'grepstr', 'grepfile'])
+TestCase = namedtuple("TestCase", ['name', 'variant', 'cmd', 'grepstr', 'grepfile', 'simlog', 'altcommand'], defaults=[None]) # applies the None default to altcommand
```
jordancarlin: What is the difference between `grepfile` and the new `simlog`? Looks like they are always set to the same thing.

jordancarlin (commented):
The new approach looks great! I assume you didn’t mean to check in actual_embench_results? That probably needs to be added to the gitignore.

PR #1396, opened 2025-04-30
src/generic/lzc.sv: rewrite using recursive structure

infinitymdm:
Rewrite lzc.sv using recursion and an explicit generate block. This avoids a while loop that is unsynthesizable with yosys.

### DIFF
--- a/src/generic/lzc.sv
+++ b/src/generic/lzc.sv
@@ -1,39 +1,53 @@
 ///////////////////////////////////////////
 //
 // Written: me@KatherineParry.com
-// Modified: 7/5/2022
+// Modified: 01/23/2025 by marcus@infinitymdm.dev
 //
 // Purpose: Leading Zero Counter
-// 
+//
 // A component of the CORE-V-WALLY configurable RISC-V project.
 // https://github.com/openhwgroup/cvw
-// 
+//
 // Copyright (C) 2021-23 Harvey Mudd College & Oklahoma State University
 //
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
 //
-// Licensed under the Solderpad Hardware License v 2.1 (the “License”); you may not use this file 
-// except in compliance with the License, or, at your option, the Apache License version 2.0. You 
+// Licensed under the Solderpad Hardware License v 2.1 (the “License”); you may not use this file
+// except in compliance with the License, or, at your option, the Apache License version 2.0. You
 // may obtain a copy of the License at
 //
 // https://solderpad.org/licenses/SHL-2.1/
 //
-// Unless required by applicable law or agreed to in writing, any work distributed under the 
-// License is distributed on an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, 
-// either express or implied. See the License for the specific language governing permissions 
+// Unless required by applicable law or agreed to in writing, any work distributed under the
+// License is distributed on an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+// either express or implied. See the License for the specific language governing permissions
 // and limitations under the License.
 ////////////////////////////////////////////////////////////////////////////////////////////////
 
-module lzc #(parameter WIDTH = 1) (
-  input  logic [WIDTH-1:0]            num,    // number to count the leading zeroes of
-  output logic [$clog2(WIDTH+1)-1:0]  ZeroCnt // the number of leading zeroes
+module lzc #(
+    parameter WIDTH = 1,
+    parameter CBITS = $clog2(WIDTH+1)
+) (
+  input  logic [WIDTH-1:0]  num,    // number to count the leading zeroes of
+  output logic [CBITS-1:0]    ZeroCnt // the number of leading zeroes
 );
 
-  integer i;
-  
-  always_comb begin
-    i = 0;
-    while ((i < WIDTH) & ~num[WIDTH-1-i]) i = i+1;  // search for leading one
-    ZeroCnt = i[$clog2(WIDTH+1)-1:0];
-  end
+    // Use a recursive binary tree structure to avoid unsynthesizable loops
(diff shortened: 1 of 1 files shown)


rosethompson:
What is the yosys error message?

infinitymdm:
After some discussion with @rosethompson, we're pretty sure this is a bug with yosys and should be fixed over there. I'll link the appropriate issue here once I've opened it.

davidharrishmc:
Can we close this for now?

stineje:
No idea what this was or the issue as DC is our EDA tool of choice - I would close.

infinitymdm:
This was related to getting wally working with siliconcompiler. There are a handful of small tweaks there, but this is the only one I thought might be useful upstream.

Rose and I discussed leaving it open for visibility, so others trying to use yosys with wally could find it a bit more easily. It can be closed if you'd prefer.

stineje:
Marcus,

Thanks for the post.  If you can close it, it would be great.  Knowing what I know from the yosys developers, I think that tool will be changed soon.  Take care.

All my best,

James

infinitymdm:
@stineje discovered that my implementation doesn't work when `WIDTH` is not a power of 2. This easy to verify by changing `VEC_LEN` in the testbench above.

infinitymdm:
At this point I've tested with all the FPGA-focused tools. It's a little more annoying to develop a "minimal" setup for DC and Genus, as they'll both require tech libraries. I'll keep working on this tomorrow and see where we get, but I'm comfortable marking this ready for review.

Please take a look at the flows I've thrown together for quartus and vivado. I have little experience with these tools, and these flows are intentionally oversimple; just enough to reproduce the errors that motivated this change. If you have any suggestions for improvements I'm all ears.

davidharrishmc:
Can you run DC in the OSU environment?  All of the TSMC libraries should be installed.

infinitymdm:
> Can you run DC in the OSU environment? All of the TSMC libraries should be installed.

I did end up setting it up that way. I would have preferred to use something a bit easier to verify externally though. I could probably set it up with sky130a from Open PDKs, but I don't know if that's a worthwhile use of my time just to verify this module; let me know if you think otherwise.

Anyways, I now have synthesis results with all the tools. Results are summarized here: https://github.com/infinitymdm/cvw/tree/lzc_tests/src/generic/test_lzc

Interestingly, I was unable to replicate the error reported on the prior design with Cadence tools. My understanding is that Jasper interacts with Genus for RTL synthesis, so in theory this should be a valid way to test. Of course I'm using Genus v21.16 (as opposed to Cadence Jasper v2024.12, where the error was corroborated). I unfortunately do not have access to Jasper or a newer version of Genus to test.

infinitymdm:
Here's a quick summary of the DC PPA results (with TSMC libs):

| Design                  | Total Comb. Area | Comb. Power | num -> ZeroCnt delay |
| ----------------------- | -----------------| ----------- | -------------------- |
| Prior                   | 19.404           | 3.104e-3 mW | 0.17                 |
| Recursive (unoptimized) | 13.230           | 1.809e-3 mW | 0.19                 |
| Recursive (optimized)   | 4.536            | 1.104e-3 mW | 0.02                 |

It rather looks like we have a clear winner here. That last row of numbers literally made me spit out my coffee.

As stated previously, I'm not 100% confident in these results, so please review the logs available here: https://github.com/infinitymdm/cvw/tree/lzc_tests/src/generic/test_lzc/logs

infinitymdm:
Genus results are a little different, and more along the lines of what I was expecting:

| Design                  | Total Comb. Area | Comb. Power | num -> ZeroCnt delay |
| ----------------------- | ---------------- | ----------- | -------------------- |
| Prior                   | 92.527           | 4.621e-7 W  | 132 ps |
| Recursive (unoptimized) | 92.527           | 4.815e-7 W  | 125 ps |
| Recursive (optimized)   | 96.032           | 4.696e-7 W  | 118 ps |

I'm a bit surprised that the area for the optimized design is larger, but that's what the tool says. Reports are visible in the logs as with DC.

stineje:
Marcus,

Great job. I think the next step is to push the performance (timing) harder to see whether the synthesis flow is stopping at some indeterminate point. We’ll likely need to adapt the script to work with non–DC-based synthesis, since these tools can be opaque and it’s not always clear how they make decisions.

In CVW we built a script that sweeps through multiple timing targets until the flow fails, and that approach gives a much clearer picture of what’s actually happening. I think doing something similar here for PPA analysis would really help us understand whether this behavior matches what we’ve seen before or if something has changed.

Thanks again—nice work. Take care.

All my best,

James

On Nov 21, 2025, at 10:44 AM, Marcus Mellor ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/35574527?s=20&v=4]infinitymdm left a comment (openhwgroup/cvw#1396)<https://github.com/openhwgroup/cvw/pull/1396#issuecomment-3563831398>

Genus results are a little different, and more along the lines of what I was expecting:

Design  Total Comb. Area        Comb. Power     num -> ZeroCnt delay
Prior   92.527  4.621e-7 W      132 ps
Recursive (unoptimized) 92.527  4.815e-7 W      125 ps
Recursive (optimized)   96.032  4.696e-7 W      118 ps

I'm a bit surprised that the area for the optimized design is larger, but that's what the tool says. Reports are visible in the logs as with DC.

davidharrishmc:
It’s worth looking at the netlists for an 8-bit design if that is big
enough to reproduce the strange results.

davidharrishmc:
Any news?  It may be hard to draw conclusions about timing from Wally as a whole.

infinitymdm:
No news, unfortunately. I've been having issues getting the wally rv64i synthesis to run to completion; my process keeps getting killed. Haven't dug into why yet.

I'll switch tack and try the 8-bit and 64-bit syntheses you've mentioned. Will put together a small clocked top module so I can set up the timing constraints correctly.

davidharrishmc:
Any news?

Issue #1263, opened 2025-02-04
New Buildroot FPGA bug


rosethompson:
The new Buildroot successfully boots on the vcu108 board and is able to run commands; however, it cannot switch to zsh.  zsh locks up the terminal and the software appears to be stuck in a trap storm (probably interrupt) as the SPI device has a pending interrupt which is never acknowledged. I don't have (m/scause) logged in the ILA.

jordancarlin:
Exciting that the new version is running on the FPGA! Strange issue with the interrupts. I wonder if it's configuration related with new version of SBI?

rosethompson:
I understand what is going on now. The above analysis has an error, please ignore.

This bug occurs when a misaligned memory access spans two pages and both cause DTLB misses and the second causes a D cache miss.  After the HPTW writes the first DTLB fill, the HPTW drives PAdrM with HPTWAdr; however, the LSU thinks the HPTW is done and allows the D cache to re-issue the original memory request.  In this manifestation a store double causes a memory fetch.  This fetch is then incorrectly writes into the cache's SRAM at PAdrM.  The HPTW begins walking the second page and gets a corrupted PTE.

The LSU prevents the D cache from re-issuing the original memory request by asserting SelHPTW.  The problem is after the first walk the HPTW FSM returns to IDLE lowering SelHPTW.  I think the solution is to find a simple way to continue to assert SelHPTW during this special case.

davidharrishmc:
This is great!  Is there a way to get cases like this into our functional verification?

rosethompson:
Partial good news. I created a test program that triggers this specific bug and a single line RTL solution that fixes the bug in simulation. However, the zsh still fails, but for a different reason. The first page table walk and first d cache access are correct.  The second page table walk gets to the last level of the table and reads a PTE of 0.  I'm not sure why this happens yet.

rosethompson:
I understand the second error now. The stval register contains the incorrect value.  We are recording the base virtual address of the store.  However, it is the spilled half which is missing from the page table. stval needs to be written with the address of the part of the access which causes the fault.  Quoted from  section 3.1 page 45 of the privilege spec 

> If mtval is written with a nonzero value when an instruction access-fault or page-fault exception
> occurs on a hart with variable-length instructions, then mtval will contain the virtual address of the
> portion of the instruction that caused the fault, while mepc will point to the beginning of the
> instruction.

There are probably similar bugs we haven't considered with xepc, and xcause.

[...]

PR #1404, opened 2025-05-07
Fixes two bugs with Issue #1263

rosethompson:
Bug 1:  Misaligned address spanning two pages corrupted the page table. After the first HPTW walk, the first part of the original memory request is made.  Then the align module issues the second half of the access.  The bug occurred inside the align module.  As the align module transitioned into the Spill state it increments PAdrM to the next cache line boundary.  However because the second PTE is not present in the TLB, the aligner should not select the incremented address.    The solution is to block selecting the spill increment during a walk.

Bug 2:  A misaligned page fault updates xtval with the portion of the instruction's address which causes the fault, not the base address.  In this case the offset address was 0x...fff with a store double.  The first page hits, but the second does not.  Wally was incorrectly loading the effective virtual address of the store double into stval rather than the spilled virtual address 0x...fff+1


rosethompson:
A few notes about this bug.

1. My regression test only checks for Bug 1.
2. Bug 2 still needs a test.
3. Misaligned instruction page faults will have this issue also and should be fixed in another pr or this one.

rosethompson:
Bug 2 now has a test that found an additional error with misaligned page faults.  The value of mtval needs to be the first byte address of the part of the access which causes the fault.  This is different than the misaligned address used by the LSU.

davidharrishmc:
Let’s figure out why these aren’t being caught by our functional coverage tests.

Issue #1406, opened 2025-05-07
Misalgined Instruction Pagefault Crossing Virtual pages Bug


rosethompson:
Similar to Issue #1263, misaligned instruction page faults do not set the correct xtval.  I have a proposed fix, but it adds 3 XLEN registers to pipeline PCSpillF to the Memory stage.  I also wrote a regression test to drive Wally into this state; however, I am having difficulty getting the program to terminate gracefully after the exception.  It will likely require a special trap handler different from the other coverage tests.  The test attempts to execute an instruction from a page with an invalid PTE.  The trap handler needs to jump somewhere other than the next instruction or it needs to update the PTE with a valid entry.

davidharrishmc:
Shouldn’t a page fault on a page with an invalid PTE just terminate the process?

PR #1411, opened 2025-05-08
Fixes Misalgined Instruction Page Fault 

rosethompson:
Resolved issue #1406. Also adds a test to prove the test works.

The RTL change adds 3 1 bit pipeline registers and an XLEN incrementer and XLEN mux.  PCM is incremented by +2 when an instruction address spill across two virtual pages. If the second page causes a fault xtval is written with the address of the second half of the instruction.

This PR also includes minor simplifications of the ebufsmarb.


rosethompson:
I ended up duplicating the WALLY-init-lib.h code to add a special handler for instruction page faults.  I didn't want to constrain the other tests to use the same policy for instruction page faults so I copied the code to another file.  We can merge the two if we don't anticipate handling other instruction page faults in more sophisticated ways (just terminating the test).

davidharrishmc:
Generally our trap handler just logs a trap and returns to the next instruction after the fault.

PR #1415, opened 2025-05-10
Functional coverage from Sail + Update Sail version

jordancarlin:
- Update to newer version of Sail and add (mostly default) config files for it. These will likely need additional updates to actually match our configs. The only change I made from the default was enabling misaligned support for rv64.
- Update custom version of RISCOF plugin to run functional coverage based on cvw-arch-verif coverpoints. Most of this will need to be upstreamed eventually.
- Updates to makefile target in `cvw/tests/riscof` directory. Running `make cvw-riscv-arch-test` will now collect coverage on tests from cvw-riscv-arch-test in addition to running them on Spike and Sail


davidharrishmc:
test.vh
	Please make an issue for this, and mention it in the comment
// TODO: Sail misaligned is turned on but these still mismatch. Need to investigate
// "rv64i_m/privilege/src/misalign-ld-01.S",

.json
Sv57 isn’t supported yet.
Maybe rv32 shouldn’t support Sv39+ and rv64 shouldn’t support Sv32?

jordancarlin:
Updated the Sail config files to match our specific implementations instead of just having the (mostly) generic versions. This means many of the wally-riscv-arch-test tests that previously had hand-written signatures can now use Sail for the signature.

Also fixed a config selection option and now the riscv-arch-test misaligned access tests match between Spike, Sail, and Wally, so they are included in regression.

PR #1427, opened 2025-05-22
Changes in wally, sail, and spike config to support vector verification

georgiatai:
Add changes to config files in riscof and in config to complete the work flow for automating vector test generation and verification

### DIFF
--- a/config/rv32gc/coverage.svh
+++ b/config/rv32gc/coverage.svh
@@ -17,6 +17,7 @@
 `define M_COVERAGE
 `define F_COVERAGE
 `define D_COVERAGE
+`define VX8_COVERAGE
 `define ZBA_COVERAGE
 `define ZBB_COVERAGE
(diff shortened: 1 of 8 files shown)


georgiatai:
That's fair. I have notes on what I changed too so I can just update those who need it. Will close the PR then!

davidharrishmc:
It's too hard to keep a separate branch for a project this big with multiple people.  The changes don't break Wally, but just add capabilities it doesn't officially support yet.  I'm going to merge this.

jordancarlin:
> It's too hard to keep a separate branch for a project this big with multiple people. The changes don't break Wally, but just add capabilities it doesn't officially support yet. I'm going to merge this.

This will probably break regression when we start running the priv tests on Wally because different CSRs will be accessible. Fine for now, but will need to revert when that starts happening.

PR #1434, opened 2025-05-27
Updated Riscof Makefile so that it doesn't give invalid reports

jacassidy:


jacassidy:
Outside of the make clean, would it make sense to make cvw-arch% dependent on the config$.svh file respectively and could that solve this issue?

davidharrishmc:
Looks reasonable to me, but I'll defer to Jordan on the Makefile.

jacassidy:
@jordancarlin are you okay with this merge

Issue #1419, opened 2025-05-13
fpu.S coverage test objdump fails with binutils 2.44


jordancarlin:
With the latest version of binutils (2.44), generating an objdump for the `fpu.S` coverage test fails with an `Aborted (core dumped)` message. All other tests are fine with the latest version of objdump.

```
$ riscv64-unknown-elf-objdump -S -D fpu.elf

fpu.elf:     file format elf64-littleriscv


Disassembly of section .text:

0000000080000000 <rvtest_entry_point>:

.section .text.init
.global rvtest_entry_point

rvtest_entry_point:
    la sp, topofstack       # Initialize stack pointer (not used)
    80000000:   00005117                auipc   sp,0x5
    80000004:   20010113                addi    sp,sp,512 # 80005200 <topofstack>

    # Set up interrupts
    la t0, trap_handler
    80000008:   00000297                auipc   t0,0x0
    8000000c:   04828293                addi    t0,t0,72 # 80000050 <trap_handler>
    csrw mtvec, t0      # Initialize MTVEC to trap_handler
    80000010:   30529073                csrw    mtvec,t0
    csrw mideleg, zero  # Don't delegate interrupts
    80000014:   30301073                csrw    mideleg,zero
    csrw medeleg, zero  # Don't delegate exceptions
    80000018:   30201073                csrw    medeleg,zero
#    li t0, -1           # set mtimecmp to biggest number so it doesnt interrupt again
#    li t1, 0x02004000   # MTIMECMP in CLINT
#    sd t0, 0(t1)      
    li t0, 0x80         
    8000001c:   08000293                li      t0,128
#    li t0, 0x00         
    csrw mie, t0        # Enable machine timer interrupt
    80000020:   30429073                csrw    mie,t0
    la t0, topoftrapstack 
    80000024:   00005297                auipc   t0,0x5
    80000028:   3dc28293                addi    t0,t0,988 # 80005400 <topoftrapstack>
    csrw mscratch, t0   # MSCRATCH holds trap stack pointer
    8000002c:   34029073                csrw    mscratch,t0
    csrsi mstatus, 0x8  # Turn on mstatus.MIE global interrupt enable
    80000030:   30046073                csrsi   mstatus,8
    # set up PMP so user and supervisor mode can access full address space
    csrw pmpcfg0, 0xF   # configure PMP0 to TOR RWX
    80000034:   3a07d073                csrwi   pmpcfg0,15
    li t0, 0xFFFFFFFF   
[...]

PR #1448, opened 2025-06-04
Installation overhaul

jordancarlin:
Installation CI is finally succeeding again after a few months of consistent failures. Several significant changes here:

- I broke the `wally-tool-chain-install.sh` script into individual scripts for each tool (`sail-install.sh`, `spike-install.sh`, etc.).  `wally-tool-chain-install.sh` calls each of these so it can still be used to install everything as before, but if you just want to update one tool without touching anything else you can run the individual scripts. These smaller scripts are also much easier to follow and hopefully more maintainable.
- Each tool is now pinned to a specific version/release/hash so that the same version is always installed on all systems. This should help avoid some of the issues we’ve seen with reproducibility and will be helpful when we create the 1.0 tag of the repo for the book so everyone gets versions of the tools that work with the current scripts in the repo.
  - This also means rerunning the installation script will only update tools that we have explicitly changed the pinned version of. 
  - The desired versions are stored in variables at the top of each of the individual installation scripts, so should be easy to update when we want to move to newer versions of any tools.
- The Linux Makefile was updated to only disassemble code sections instead of all contents in the `fw_jump.elf` file. This is slightly unfortunate, but is necessary to work around the bug in objdump that we already had to deal with in some of our tests. Reverting to an earlier version of binutils is not an option because of other issues in the riscv-gnu-toolchain repo, so this is the best I could come up with. I went through and verified that all examples of the OpenSBI code that we discuss in the book are in the section that still gets disassembled, so I don't think this will have a significant impact.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -149,7 +149,7 @@ jobs:
       # Update setup.sh if using a custom $RISCV path
       - name: Update setup.sh
         if: ${{ matrix.riscv_path != null }}
-        run: sed -i 's,/opt/riscv,${{ matrix.riscv_path }},g' setup.sh
+        run: sed -i 's,~/riscv,${{ matrix.riscv_path }},g' setup.sh
       # Upload installation logs for debugging
       - name: Upload Installation Logs
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


rosethompson:
What is the objdump bug?  Could we also output the data sections?  I find these sections useful when doing FPGA debugging.  OpenSBI is less important than Linux and buildroot so if it only impacts OpenSBI I am probably OK with this.

jordancarlin:
@rosethompson We were only disassembling the whole file for `fw_jump.elf`. All of the other files (including the `vmlinux` binary) were only disassembling the code regions.

It's the same issue we ran into here (https://github.com/openhwgroup/cvw/issues/1419) and patched for one of the fpu coverage tests here (https://github.com/openhwgroup/cvw/pull/1420).

PR #1454, opened 2025-06-07
Update Sail

jordancarlin:
Update Sail version and config files. Fixes vector log parsing issue.


jordancarlin:
Skipping CI since it will fail because of the PMP test issues anyway. This can be merged though.

PR #1456, opened 2025-06-11
Temporarily return G=0 so riscof can build PMP ACTs

davidharrishmc:


davidharrishmc:
Also fixed Wally tracer for interrupts 
https://github.com/openhwgroup/cvw-arch-verif/issues/670

PR #1459, opened 2025-06-18
VC Spyglass Lint scripts

stineje:
Add VC Spyglass bash script lint_spyglass.sh.  Unfortunately, due to a bug in Spyglass that was correlated with Synopsys, the waivers have to be a separate file called waiver.tcl.  There is a workaround but it's very messy, so just left as is.  The script can run multiple configs and writes all data to the lint-spyglass-reports subdirectory that it makes in the current directory.  These files can also be pulled up with a HTML browser or Spyglass itself.


bin/lint-spyglass.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+###############################################################################
+# Script Name: lint_spyglass.sh
+# Author: james.stine@okstate.edu 11 June 2025
+# Description:
```
jordancarlin, line 7: Missing license in all files.

jordancarlin (commented):
Should we create a spyglass directory somewhere to put the cvw.prj and waivers.tcl files? They aren’t executable scripts so don’t really belong in bin. We could also have the report get generated in that directory instead of the current directory.

Also, does something need to be added to site-setup.sh to add Spyglass to the PATH?

bin/lint-spyglass.sh
stineje: Sorry about that - will fix.  Great catch!

stineje: Thanks for the comment; thought about that originally but didn't change it.  I moved to synthDC/spyglass for cvw.prj and waivers.tcl as seems like a good place.  Thanks for the suggestions!!!!

bin/lint-spyglass.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+# OPTIONAL: choose one of: lint/lint_rtl or lint/lint_rtl_enhanced
+GOAL="lint/lint_rtl"
+
+# List of configurations (add configurations for linting)
+configs=(rv32e rv32i rv64i rv64gc)
+
+# Set current dir to make sure it writes to the current dir
+CUR_DIR=$(pwd)
```
jordancarlin: Instead of `CUR_DIR`, which will change depending on where you call the script from, should we use `SPYGLASS_DIR=$WALLY/synthDC/spyglass`?

bin/lint-spyglass.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Base project file for Spyglass
+TEMPLATE_PRJ="$WALLY/synthDC/spyglass/cvw.prj"
+
+# Clean out the lint-synopsys directory (if it exists)
+if [ -d "$CUR_DIR/lint-spyglass-reports" ]; then
+    echo "Cleaning lint-spyglass-reports directory..."
+    rm -rf "$CUR_DIR/lint-spyglass-reports"/*
```
jordancarlin: You can unconditionally delete this to simplify things a bit. There’s no error if it doesn’t already exist.

bin/lint-spyglass.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Error: WALLY environment variable is not set."
+    echo "Please make sure you source your setup before running this script."
+    exit 1
+fi
+
+# OPTIONAL: choose one of: lint/lint_rtl or lint/lint_rtl_enhanced
+GOAL="lint/lint_rtl"
+
+# List of configurations (add configurations for linting)
+configs=(rv32e rv32i rv64i rv64gc)
```
jordancarlin: Should the default set be the full list of default configs? rv32e, rv32i, rv32imc, rv32gc, rv64i, rv64gc

Also, might be nice if this was an option to set on the command line instead on needing to edit the script.

bin/lint-spyglass.sh
```diff
@@ -0,0 +1,72 @@
...
+
+    # Output project file
+    CONFIG_PRJ="$CUR_DIR/cvw_${config}.prj"
+
+    # Replace WALLYVER with current config and save to new .prj
+    # Also replaces path for Tcl so can incorporate everything correctly
+    sed -e "s|\$WALLY|$WALLY|g" \
+	-e "s|WALLYVER|$config|g" \
+	-e "s|read_file -type awl waivers.tcl|read_file -type awl $WALLY/synthDC/spyglass/waivers.tcl|g" \
+	-e "s|set_option projectwdir lint-spyglass/|set_option projectwdir ${CUR_DIR}/lint-spyglass/|g" \
```
jordancarlin: Should the workdir be added to the .gitignore?

site-setup.sh
```diff
@@ -27,6 +27,9 @@ export BREKER_HOME=/cad/breker/TREK                                 # Change thi
 # Tools
 # Questa and Synopsys
 export PATH=$QUESTA_HOME/bin:$DC_HOME/bin:$VCS_HOME/bin:$PATH
+# Synopsys Spyglass
+export SPYGLASS_HOME=/cad/synopsys/spyglass/W-2024.09-SP2-2/SPYGLASS_HOME
```
jordancarlin: Let’s move this up to be with the other _HOME variables. Also, does `$SPYGLASS_HOME/bin` (or something like that) need to be added to the path?

Would also be good to match the other HOME variables and set this to `/cad/synopsys/SPYGLASS` and then it can be further customized on individual servers.

bin/lint-spyglass.sh
stineje: done

bin/lint-spyglass.sh
stineje: done

bin/lint-spyglass.sh
stineje: done

bin/lint-spyglass.sh
stineje: done

site-setup.sh
stineje: The original reason for keeping it below here is that it expanded the comment line width.  I agree with you here, so moved as suggested.

PR #1466, opened 2025-06-25
Adding Missing extensions

Marina-Bellido:
In the Makefile: I added a command to also run privileged tests in the privileged folder
Added missing extensions: Zmmul and Zfhmin


davidharrishmc:
Is it necessary to add Zmmul and Zfhmin to the yaml?

PR #1494, opened 2025-07-30
Update Sail to new single binary version

jordancarlin:


tests/riscof/sail_cSim/rv64gc.json
```diff
@@ -104,6 +112,9 @@
     "Zabha": {
       "supported": false
     },
+    "Zacas": {
+      "supported": false
+    },
     "Zalrsc": {
       "supported": false
     },
```
davidharrishmc, line 120: What does it mean for A to be supported but Zalrsc to not be supported?  Should Zaamo and Zalrsc both be supported?

jordancarlin: This is how we’ve had it setup for a while. A enables both sets of instructions and the misa.A bit while Zalrsc and Zaamo each enable their respective subset of instructions without the MISA bit. Since our MISA is read only, Zalrsc and Zaamo have no impact if A is already enabled.

davidharrishmc:
Merged this, but still wondering if we should set Zalrsc and Zaamo too.

PR #1502, opened 2025-08-08
Introduce Wally development containers

doganulus:
This PR introduces two development containers, tagged as `builder` and `devel`, each including all dependencies required to develop and test the Wally processor.

- **`builder`** images intended for CI workflows
- **`devel`** images intended for end-users

These containers will serve as a quick start for setting up a development environment and testing the Wally processor. The container build workflow produces images for both `amd64` and `arm64` platforms.

Currently marked as a draft, as the ARM build depends on an upcoming upstream release from the SAIL project. Container images without `sail` installed are available temporarily here for testing/demo purposes:  
- https://github.com/doganulus/cvw/pkgs/container/wally

Note that the current base image is Debian 12. We can discuss this choice along with what should be included in the containers.


.github/workflows/container-release.yml
```diff
@@ -0,0 +1,178 @@
+name: Container Release
+on:
+  push:
+    paths:
+      - containers/wally-devel/**
+      - .github/workflows/container-release.yml # Self-trigger
```
jordancarlin, line 10: Changes to the installation scripts should also trigger this to run.

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,178 @@
...
+  workflow_dispatch:
+    inputs:
+      publish:
```
jordancarlin, line 13: It would be nice to have an option to set a different version in the workflow dispatch inputs so that we can keep tagged versions of the images around (like the versions of the tools corresponding to the upcoming 1.0 release of the repo).

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,178 @@
...
+        description: "Publish to the container registry"
+        required: false
+        default: true
+        type: boolean
+
+env:
+  REGISTRY: ghcr.io/doganulus
```
jordancarlin, line 20: Will need to be changed for the final version (just noting this so we don't forget before merging).

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,178 @@
...
+  IMAGE_NAME: wally
+  PODMAN_ARCH: amd64
+  PLATFORM: linux-amd64
```
jordancarlin: Should these be set as part of the matrix so they are set appropriately for arm builds?

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,178 @@
...
+        run: |
+          echo "PODMAN_ARCH=$(podman info --format='{{.Host.Arch}}')" >> $GITHUB_ENV
+          echo "PLATFORM=$(podman info --format='{{.Version.OsArch}}' | sed 's/\//-/g')" >> $GITHUB_ENV
+          echo "CONTAINER_VERSION=$(date +'%Y%m%d')" >> $GITHUB_ENV
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the GitHub Container registry
+        uses: redhat-actions/podman-login@v1
```
jordancarlin, line 72: What is the rationale for using the redhat-actions instead of the more common [docker-build-push action](https://github.com/marketplace/actions/build-and-push-docker-images)?

containers/wally-devel/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG CONTAINER_VERSION=latest
+ARG CONTAINER_BASE_IMAGE=docker.io/library/debian:12
```
jordancarlin: What about using Rocky Linux or AlmaLinux as the base image? That should allow people to mount local commercial CAD tools when running the image if desired.

containers/wally-devel/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    bash "$WALLY/bin/installation/riscv-gnu-toolchain-install.sh" --clean
+
+RUN --mount=type=bind,source=./bin,target=/wally/bin \
+    export WALLY=/wally && \
+    export RISCV=/opt/riscv && \
+    bash "$WALLY/bin/installation/elf2hex-install.sh" --clean && \
+    bash "$WALLY/bin/installation/qemu-install.sh" --clean && \
+    bash "$WALLY/bin/installation/spike-install.sh" --clean && \
+    # bash "$WALLY/bin/installation/sail-install.sh" --clean && \
+    bash "$WALLY/bin/installation/verilator-install.sh" --clean
```
jordancarlin: Have these been tested yet? It seems like right now the `site-setup.sh` script is not installed into `$RISCV`, which is likely to cause issues. It also doesn't run the `python-setup.sh` script, which is needed to run RISCOF.

I wonder if we would be better off running the full `$WALLY/bin/wally-tool-chain-install.sh` script. That ensures all components are installed and would probably make maintenance easier if new tools are added. If you're worried about layers, we could pre-run the `riscv-gnu-toolchain-install.sh` script so it is in a separate layer. The main installation script will skip it if it sees it is already the correct version.

jordancarlin (changes requested):
This is great! I was working on a Docker image a few months ago, but unfortunately never found time to finish it. I left a first round of detailed comments inline.

One more general question is whether we really need separate builder and devel images. While I see the value of that for some projects, I'm not sure what would really be different between them in the case of the Wally images. It is also simpler for users if there is just a single image for them to pull.

.github/workflows/container-release.yml
doganulus: This or adding tagged versions to the build matrix. Both are fine.

.github/workflows/container-release.yml
doganulus: These variables are overwritten in matrix builds. Set to an initial value to keep the linter happy.

.github/workflows/container-release.yml
doganulus: Such multi-arch builds must build in separate runners and merge their manifests. 

If I recall correctly, my reason to switch `redhat-actions` was this issue: https://github.com/docker/build-push-action/issues/1287

containers/wally-devel/Dockerfile
doganulus: Yes, this is up to the project.

Even I would like to experiment with `bootc` at some point to streamline our hardware lab management.

containers/wally-devel/Dockerfile
doganulus: Not tested yet. Some of those setup scripts can be reevaluated for container use cases. But I haven't looked up at them yet.

davidharrishmc:
I'm very excited to see this capability.  @doganulus thank you for stepping up with this.  Docker is outside my main skill set and I have questions about maintainability.  Everything in Wally has been subject to fast bit rot and I'd like to make sure Docker is maintainable if we direct users this way.  @doganulus will you / your team be willing to own this capability for a good while (say initially 3 years).  This would include preparing documentation sufficient for low-skill users, fixing problems, and providing basic support (preferably little because the documentation is good and there are no bugs, but that won't be the case immediately).  It would also be great to get this into a regular regression that would notify everyone, especially you, when some dependencies change and the container stops working.

I don't have experience to judge between builder & devel vs. just devel, but my bias is toward minimalism if devel could be sufficient.  Fewer things to maintain and for other people to have to understand.

I don't fully understand the security issues on Docker.  Two years ago, we had some concerns about whether running the cvw environment in Docker would give too much privilege to users.  Is this setup now something that a paranoid corporate IT department would comfortably allow on their system?

Regarding Debian, I'd like to understand tradeoffs.  I'm ok with it if there are compelling technical reasons to go there.  So far, everything for Wally including the textbook documentation has been on Rocky and Ubuntu.  Rocky is the free Red Hat distro that is compatible with EDA vendors including Synopsys, Cadence, Siemens, and Breker.  Ubuntu is popular with desktop users and has been sufficient to run almost all of these tools except recent Design Compiler.  We've moved our work from Ubuntu to Rocky to be able to use all the EDA.  If a user wants to synthesize the RTL in Design Compiler, does it matter whether the Docker container is running Debian or Red Hat?

doganulus:
> I don't fully understand the security issues on Docker. Two years ago, we had some concerns about whether running the cvw environment in Docker would give too much privilege to users. Is this setup now something that a paranoid corporate IT department would comfortably allow on their system?

This is caused by the fact that the Docker daemon requires root access by default, which may be _rightfully_ unwanted in shared user systems. However, containers are now a general technology, and it is possible (and preferred) to run them _rootless_ in these environments. Then, actually, a container image may be preferred by users and IT admins alike, as it is just pull and run.

> Regarding Debian, I'd like to understand tradeoffs. I'm ok with it if there are compelling technical reasons to go there.

The differences would be minimal. Currently, there is a bug in the installation scripts (a hardcoded path for `amd64` builds) regarding Ubuntu, so I switched to Debian for this demonstration. Yes, I agree that Red Hat based distros may be better due to their ecosystem.

> @doganulus will you / your team be willing to own this capability for a good while (say initially 3 years). 

We can do that, yes.

davidharrishmc:
This sounds great!

containers/wally-devel/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+ARG WALLY_CONTAINER_BUILDER_IMAGE=wally-builder
+ARG WALLY_CONTAINER_DEVEL_IMAGE=wally-devel
+
+ARG RISCV_GNU_TOOLCHAIN_VERSION=23863c2ca74e6c050f0c97e7af61f5f1776aadd1
+
+FROM ${WALLY_CONTAINER_BASE_IMAGE} as wally-builder
+ARG TARGETARCH TARGETOS TARGETPLATFORM TARGETVARIANT
+
+ARG RISCV_GNU_TOOLCHAIN_VERSION
+ENV RISCV_GNU_TOOLCHAIN_VERSION=${RISCV_GNU_TOOLCHAIN_VERSION}
```
jordancarlin, line 13: This is already set in the installation script. Any reason to set it here as well?

containers/wally-devel/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+ENV CPATH="${RISCV_INSTALL_PREFIX}/include${CPATH:+:${CPATH}}"
+ENV LD_LIBRARY_PATH="\
+        ${RISCV_INSTALL_PREFIX}/lib\
+        :${RISCV_INSTALL_PREFIX}/lib64\
+        ${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+ENV PKG_CONFIG_PATH="\
+        ${RISCV_INSTALL_PREFIX}/lib/pkgconfig\
+        :${RISCV_INSTALL_PREFIX}/lib64/pkgconfig\
+        :${RISCV_INSTALL_PREFIX}/share/pkgconfig\
+        ${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}}"
```
jordancarlin: All of these paths (plus several other variables) are set in `setup.sh` and `site-setup.sh`. What about having the `CMD` or `ENTRYPOINT` step source the `setup.sh` script so everything is set when the user enters the container?

containers/wally-devel/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN ${WALLY_PYTHON_EXECUTABLE} -m pip install --no-cache-dir \
+        reuse \
+        pre-commit
```
jordancarlin: `pre-commit` is already installed from the `requirements.txt` file.

containers/wally-devel/Dockerfile
jordancarlin: Or we could modify the `.bashrc` in the container. That would allow the file to be sourced only if it’s found.

doganulus: I am experimenting with those variables currently. Entrypoint might be a good idea. Need to check how they are overwritten when run as a `devcontainer` for `vscode`.

davidharrishmc:
This has been idle for a long time.  Closing the draft for now to simplify watching PRs.  Feel free to reopen it when it is no longer in draft form.

PR #1503, opened 2025-08-09
Seperate curl install across distros

doganulus:
This is due to RHEL-based distros coming with `curl-minimal` installed and clashing if `curl` is asked to install additionally. 

Maybe it is better to separate `curl` install across distros, assuming `curl-minimal` is sufficient for installation tasks on RHEL-based distros.

This removes a workaround in the Actions workflow.


jordancarlin (approved):
Looks like `curl-minimal` is sufficient. Thanks for the simplification. The CI failure is for an unrelated issue and the installation portion itself worked fine.

Issue #1504, opened 2025-08-09
Add installation support for new distro releases


jordancarlin:
New versions of several of our supported distros have been released recently. They need to be tested and support needs to be added to the installation script for them.

- [ ] RHEL/AlmaLinux/Rocky Linux 10
- [ ] Debian 13

PietDelaney:
I installed Debian 13 today. Hybernations works better than Ubuntu 24.04.3 LTS.
I was wondering today if you guys were interested in Debian 13. Glad to see interest.

PR #1509, opened 2025-08-16
Add support for RHEL 10 family distros

jordancarlin:
Update installation scripts and documentation to support RHEL 10. Also adds AlmaLinux 10 and Rocky Linux 10 to installation CI.

Related to #1504


jordancarlin:
This includes #1508 in it so that CI passes.

Issue #1496, opened 2025-07-30
Getting a MISMATCH when accessing invalid PTE using CBO instructions


Zain2050:
I am working on the covergroup **RV64_CBO_VM** and getting a mismatch. When I execute cbo.clean (rs1), where rs1 has VA to an invalid PTE, I get a mismatch where Wally is not faulting at the invalid PTE, but imperas is giving me a Store Page Fault.
You can recreate this using this binary 
[cbo_vm_exception.elf.gz](https://github.com/user-attachments/files/21513712/cbo_vm_exception.elf.gz)

```
# Info 671: 'refRoot/cpu', 0x0000030080000b70(main+370): Supervisor 01060613 addi    x12,x12,16
# Info   MEMX 0x30080000b70 0x80000b70 2 0613
# Info   MEMX 0x30080000b72 0x80000b72 2 0106
# Info   x12 0000000000000000 -> 0000000000000010
# Info 672: 'refRoot/cpu', 0x0000030080000b74(main+374): Supervisor 0017a00f **cbo.clean x15**
# Info   MEMX 0x30080000b74 0x80000b74 2 a00f
# Info   MEMX 0x30080000b76 0x80000b76 2 0017
# Info (OP_NCH) GlobalTime:0.000000 LocalTime:0.000007 Net:refRoot/coverpoint 0 => 4
# Info (OP_NCH) GlobalTime:0.000000 LocalTime:0.000007 Net:refRoot/coverpoint 4 => 3
# Info (OP_NCH) GlobalTime:0.000000 LocalTime:0.000007 Net:refRoot/coverpoint 3 => 0x1D
# Info   MEMRM 0x80008028 0x80008028 8 00000000000000ce L3 (refRoot/cpu/64-bit Supervisor Physical unified)
# Info (OP_NCH) GlobalTime:0.000000 LocalTime:0.000007 Net:refRoot/coverpoint 0x1D => 0xC
# Info (RISCV_PTWE) CPU 'refRoot/cpu': Page table entry V=0 [address=0x28080000400 PTEAddress=0x80008028 access=R]
# Info (OP_NCH) GlobalTime:0.000000 LocalTime:0.000007 Net:refRoot/coverpoint 0xC => 1
# Warning (RISCV_IMA) CPU 'refRoot/cpu' 0x30080000b74 0017a00f cbo.clean x15: Page fault at store/atomic memory operation address (0x28080000414)
[...]

PR #1521, opened 2025-09-10
RTL fixes for VM_CBO

Zain2050:
Closes Issue #1496


davidharrishmc:
PreUpdateDA portion looks good to me.
@rosethompson do you have time to review the other part?

Zain2050:
> You'll need to add the similar CMOpM mux to drive CMOpM='0 when SelHPTW = 1.

Done.

Zain2050:
Yes, it does. The previous changes had also resolved the problem & fixed the mismatches but this is a cleaner way to do it.

PR #1526, opened 2025-10-06
wally-riscv-arch-test now passing regression

davidharrishmc:
Disabled NA4 because it is incompatible with PMP grain > 0, fixed link.ld with chatGPT to eliminate warnings about rwx in LOAD segment

All PMP mismatches now fixed.

Hopefully this fixes nightly regression.

### DIFF
--- a/tests/riscof/sail_cSim/env/link.ld
+++ b/tests/riscof/sail_cSim/env/link.ld
@@ -1,18 +1,52 @@
 OUTPUT_ARCH( "riscv" )
 ENTRY(rvtest_entry_point)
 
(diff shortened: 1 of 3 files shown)


tests/riscof/sail_cSim/env/link.ld
jordancarlin, line 1: Are we better off just adding `-Wl,--no-warn-rwx-segments` to the compilation flags so that it doesn't produce that warning? That way we can keep a simpler linker script that people are more likely to understand. I know a number of other test suites use that flag to suppress this warning.

PR #1550, opened 2025-11-05
SKY130 install script update

stineje:
update SKY130 location as Google has removed from their website.  The library is still open-source and just moved to a different location.

### DIFF
--- a/bin/installation/skywater-lib-install.sh
+++ b/bin/installation/skywater-lib-install.sh
@@ -29,16 +29,6 @@
 
 SKYWATER_LIB_VERSION=3e7dac7af98731b59982f99df6a71e979a44bff7 # Last commit as of Nov. 5, 2025
 
-# Ensure repo always comes from NEW GitHub
-if check_tool_version $SKYWATER_LIB_VERSION; then
-    rm -rf $RISCV/cad/lib/sky130_osu_sc_t12
-    git_checkout "sky130_osu_sc_t12" "https://github.com/stineje/sky130_osu_sc_t12" "$SKYWATER_LIB_VERSION"
-    echo "$SKYWATER_LIB_VERSION" > "$RISCV"/versions/$STATUS.version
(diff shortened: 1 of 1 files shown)


bin/installation/skywater-lib-install.sh
```diff
@@ -27,7 +27,16 @@
 ## and limitations under the License.
 ################################################################################################
 
-SKYWATER_LIB_VERSION=ac90ef0c622a9377a16b5218d9da3ac4169eeaaf # Last commit as of May 30, 2025
+SKYWATER_LIB_VERSION=3e7dac7af98731b59982f99df6a71e979a44bff7 # Last commit as of Nov. 5, 2025
+
+# Ensure repo always comes from NEW GitHub
+if [ -d "$RISCV/cad/lib/sky130_osu_sc_t12/.git" ]; then
```
jordancarlin: Because this isn’t inside the `check_tool_version` if block, it will always run, even if the installation is up to date. What about unconditionally deleting the directory as the first step in the if block to avoid the git issues? It’ll only run if the version is outdated, so should be low cost and simpler.

bin/installation/skywater-lib-install.sh
```diff
@@ -27,7 +27,12 @@
...
+  rm -rf $RISCV/cad/lib/sky130_osu_sc_t12/.git
+fi
```
jordancarlin: @stineje I meant that this deletion should move to the `if check_tool_version $SKYWATER_LIB_VERSION` block so that it is only run if the version is outdated. So something like the following:

```
if check_tool_version $SKYWATER_LIB_VERSION; then
    rm -rf $RISCV/cad/lib/sky130_osu_sc_t12
    git_checkout "sky130_osu_sc_t12" "https://github.com/stineje/sky130_osu_sc_t12" "$SKYWATER_LIB_VERSION"
    echo "$SKYWATER_LIB_VERSION" > "$RISCV"/versions/$STATUS.version
    echo -e "${SUCCESS_COLOR}OSU Skywater library successfully installed!${ENDC}"
else
```

Issue #1440, opened 2025-05-31
Update testbench to exit if an incorrectly sized elf/memfile is passed to it


jordancarlin:
Currently, the testbench loads the elf/memfile blindly and attempts to run it. Wally ends up in a trap loop immediately if it is the wrong size, but it is not always immediately apparent what the cause was. The testbench should check the memfile bitwidth before running it and exit with an error message that explains what went wrong. 

Related to #1439, but both features are useful in different scenarios.

rosethompson:
Fix pending from Lamarr Olive.

PR #1552, opened 2025-11-07
Exit testbench if elf file is misaligned with DUT bit width and provide descriptive error message

Lamarr3314:
Resolving problem # 1440: [https://github.com/openhwgroup/cvw/issues/1440]

With the proposed change, if an elf file is passed in, $fread extracts the header information. elfBitWidth variable stores an integer (32/64) representing the bit width of the elf from the header. I cast the existing XLEN to an integer and store in dutBitWidth for comparison to the elfBitWidth variable.

The existing conditions for exiting the testbench checked 0 test size & no elf. Then had one if statement checking 0 test size and an else checking no elf that returned various error messages, which was redundant.

The proposed fix alters the condition for exiting the testbench. I check for 0 test size & no elf passed in or a an elf passed in and a misalignment between the elfBitWidth and dutBitWidth. Based on which condition is true, I return a descriptive error of the unsupported elf and test or the misaligned bit widths between the DUT and the elf.

### DIFF
--- a/testbench/testbench.sv
+++ b/testbench/testbench.sv
@@ -708,7 +708,7 @@ module testbench;
   // watch for problems such as lockup, reading uninitialized memory, bad configs
   watchdog #(P.XLEN, 1000000) watchdog(.clk, .reset, .TEST);  // check if PCW is stuck
   ramxdetector #(P.XLEN, P.LLEN) ramxdetector(clk, dut.core.lsu.MemRWM[1], dut.core.lsu.LSULoadAccessFaultM, dut.core.lsu.ReadDataM,
-                                      dut.core.ifu.PCM, InstrM, dut.core.lsu.IEUAdrM, InstrMName);
+                                      dut.core.ifu.PCM, InstrM, dut.core.lsu.IEUAdrM, dut.core.lsu.StallW, InstrMName);
   riscvassertions       #(P) riscvassertions();     // check assertions for a legal architectural configuration
   riscvassertions_wally #(P) riscvassertions_wally();  // check assertions for a legal microarchitectural configuration
   loggers #(P, PrintHPMCounters, I_CACHE_ADDR_LOGGER, D_CACHE_ADDR_LOGGER, BPRED_LOGGER)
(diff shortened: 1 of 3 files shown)


testbench/testbench.sv
```diff
@@ -253,11 +271,12 @@ module testbench;
         "arch32vm_sv32": if (P.VIRTMEM_SUPPORTED) tests = arch32vm_sv32;
       endcase
     end
-    if (tests.size() == 0 & ElfFile == "none") begin
-      if (tests.size() == 0) begin
+    if (tests.size() == 0 & ElfFile == "none" || ElfFile != "none" & dutBitWidth!=elfBitWidth) begin
+      if (tests.size() == 0 & ElfFile == "none") begin
         $display("TEST %s not supported in this configuration", TEST);
-      end else if(ElfFile == "none") begin
```
jordancarlin, line 259: Doesn't look like this line should have been deleted. It will now show both of these error messages even though only one of them will ever apply.
```
$display("TEST %s not supported in this configuration", TEST);
$display("ElfFile %s not found", ElfFile);
```

testbench/testbench.sv
```diff
@@ -148,6 +154,18 @@ module testbench;
     end
     //$display("TEST = %s ElfFile = %s", TEST, ElfFile);
 
+    if(ElfFile!="none")begin//If Elf File passed in, check its bit width
```
jordancarlin: Clean up formatting (global comment throughout, only noting it here). There should be spaces surrounding comparisons (` == `, ` != `, etc.) and after closing parentheses.
```suggestion
    if(ElfFile != "none") begin // If Elf File passed in, check its bit width
```

jordancarlin (changes requested):
This will be a great improvement! I left a few comments below.

testbench/testbench.sv
```diff
@@ -148,6 +154,18 @@ module testbench;
...
+      readBytes = $fread(header, elfFD);
+      $fclose(elfFD);
+
+      if(header[4] ==1)begin
+        elfBitWidth = 32;
+      end else if(header[4] == 2)begin
+        elfBitWidth = 64;
+      end
+      dutBitWidth = integer'(P.XLEN);
+    end
```
jordancarlin: I wonder if it would be simpler to put the termination here instead of down at line 272. Then you could directly compare `header[4]` and `P.XLEN` and terminate the simulation if they mismatch.

addins/cvw-arch-verif
jordancarlin, line 1: Please review what changes are included in this PR. You’re adding 2 submodules that I don’t think you intended to.

PR #1575, opened 2025-12-03
Adjusted coremark termination to use tohost 

davidharrishmc:
and infinite loop more robustly

### DIFF
--- a/benchmarks/coremark/riscv64-baremetal/crt.S
+++ b/benchmarks/coremark/riscv64-baremetal/crt.S
@@ -59,7 +59,7 @@ _start:
 #else
   bltz t0, 1f
 #endif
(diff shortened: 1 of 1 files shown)


benchmarks/coremark/riscv64-baremetal/crt.S
```diff
@@ -59,10 +59,13 @@ _start:
 #else
   bltz t0, 1f
 #endif
-2:
+2: # store 000001 to tohost to terminate with HTIF
```
jordancarlin: Why 6 digits in the comment value? It should either be just `1` or the full number of 0s that we are storing.

davidharrishmc: Update.

PR #1581, opened 2025-12-10
refactored CSR read logic to match csrm.sv

davidharrishmc:


src/privileged/csrc.sv
```diff
@@ -150,12 +150,13 @@ module csrc  import cvw::*;  #(parameter cvw_t P) (
       end
   end
 
-  // Read Counters, or cause excepiton if insufficient privilege in light of COUNTEREN flags
+  // Read Counters, or cause exception if insufficient privilege in light of COUNTEREN flags
```
jordancarlin, line 153: This reminds me. We have a spellcheck pre-commit hook on riscv-arch-test that seems to be working pretty well. Might be worth adding to cvw too.

PR #1582, opened 2025-12-11
Spelling fixes

jordancarlin:
Spellchecker had too many false positives to enable because of strangely named signals (unlike riscv-arch-test), but it still caught a handful of typos.


davidharrishmc:
Lgtm. Failed a check?

PR #1584, opened 2025-12-11
RVVILogger Module to RVVI TEXT format for coverage

SadhviNarayanan:
This is a very initial, foundational version of the RVVI logger module, which uses the RVVI Trace to extract internal hart signals and write them out to an RVVI TEXT file in the format it requires.
Note: the output has not been validated with the Python script, I just wanted to get some initial feedback on the module to see if it was even progressing in the right direction.

### DIFF
--- a/testbench/common/rvviTextLogger.sv
+++ b/testbench/common/rvviTextLogger.sv
@@ -12,11 +12,9 @@
 ///////////////////////////////////////////
 
 module rvviTextLogger #(
-    parameter string FILENAME = "logs/rvvi_trace.txt", // TODO: not sure what default should be
-    parameter string VENDOR_NAME = "Placeholdr",
+    parameter string VENDOR_NAME = "cvw",
     parameter int VENDOR_MAJOR = 1,
-    parameter int VENDOR_MINOR = 0,
-    parameter bit COMPRESSED_WIDTH_16 = 1  // Detect compressed instructions by width
+    parameter int VENDOR_MINOR = 0
 ) (
     rvviTrace rvvi  // Only input for now - RVVI-TRACE interface
 );
@@ -29,8 +27,9 @@ module rvviTextLogger #(
     localparam int NHART = rvvi.NHART;
     localparam int RETIRE = rvvi.RETIRE;
 
-    // need someway handle the file
+    // File handling
     integer file;
+    string filename;
 
     // Track current hart (for general processors, multi-core)
     int current_hart = 0;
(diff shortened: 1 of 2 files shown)


testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+// Works with any RISC-V processor implementing RVVI-TRACE
+//
+// Converts the rvviTrace interface to RVVI-TEXT format
+// for use with coverage tools, reference models, and analysis
+//
+// Fully parameterized - no processor-specific dependencies
+///////////////////////////////////////////
+
+module rvviTextLogger #(
+    parameter string FILENAME = "logs/rvvi_trace.txt", // TODO: not sure what default should be
```
jordancarlin: This should probably be a plusarg instead of a parameter so it can be specified at runtime.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+    parameter string VENDOR_NAME = "Placeholdr",
```
jordancarlin:
```suggestion
    parameter string VENDOR_NAME = "cvw",
```

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+    parameter int VENDOR_MAJOR = 1,
+    parameter int VENDOR_MINOR = 0,
+    parameter bit COMPRESSED_WIDTH_16 = 1  // Detect compressed instructions by width
```
jordancarlin: What does this parameter mean and why is it necessary?

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+    // NET CHANGES - Track External Signal Transitions (may not be needed or correct)
+    // essentially, if something externallly changes (like interrupt) we pull from the rvviTrace interface
+
+    // net_pop() function will retrieve net changes from the standard RVVI-TRACE API interface
+    string net_name;
+    logic [XLEN-1:0] net_value;
+
+    always_ff @(posedge rvvi.clk) begin
+        // Pop all available NET changes
+        while (rvvi.net_pop(net_name, net_value)) begin
```
jordancarlin: I don’t think this will work. If you pop things off the list of net changes they won’t be visible to the any other consumers of the RVVI trace. This would break lockstep.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+        // RET or TRAP
+        if (rvvi.trap[retire_slot][hart]) begin
+            $fwrite(file, "TRAP ");
+        end else begin
+            $fwrite(file, "RET ");
+        end
+
+        // PC (formatted based on XLEN)
+        if (XLEN == 32) begin
+            $fwrite(file, " %5h ", rvvi.pc_rdata[retire_slot][hart][31:0]);
```
jordancarlin: Why is [31:0] needed? Isn’t that the full size of the signal? Global comment lots of places.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+        end else if (XLEN == 64) begin
+            $fwrite(file, "%6h ", rvvi.pc_rdata[retire_slot][hart][63:0]);
+        end else begin
```
jordancarlin: Why is this else case needed?

testbench/common/rvviTextLogger.sv
jordancarlin: Why 5? Shouldn’t this be %04h? There are a lot of these format specifiers that seem off.

testbench/common/rvviTextLogger.sv
jordancarlin: Global comment. A lot of `if` blocks seem to have unnecessary else cases.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+    task automatic writeGPRChanges(int hart, int retire_slot);
+        string reg_name;
+
+        // 32 registers to look at
+        for (int i = 0; i < 32; i++) begin
+            if (rvvi.x_wb[retire_slot][hart][i]) begin
+                // Get register name
+                reg_name = getGPRName(i);
+
+                // Format: X <index> <value>
```
jordancarlin: This comment doesn’t match what you’re actually printing.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+                if (rvvi.f_wb[retire_slot][hart][i]) begin
+                    // Get register name
+                    reg_name = getFPRName(i);
+
+                    // Format: F <index> <value>
+                    if (FLEN == 32) begin
+                        $fwrite(file, "F %2d '%s' %08h ", i, reg_name, rvvi.f_wdata[retire_slot][hart][i][31:0]);
+                    end else if (FLEN == 64) begin
+                        $fwrite(file, "F %2d '%s' %016h ", i, reg_name, rvvi.f_wdata[retire_slot][hart][i][63:0]);
+                    end else if (FLEN == 128) begin
```
jordancarlin, line 250: Is there any way to use FLEN in the format specifiers so we don’t need all of these if/else cases? Same applies to XLEN elsewhere.

testbench/common/rvviTextLogger.sv
```diff
@@ -0,0 +1,376 @@
...
+            23: return "s7";
+            24: return "s8";
+            25: return "s9";
+            26: return "s10";
+            27: return "s11";
+            28: return "t3";
+            29: return "t4";
+            30: return "t5";
+            31: return "t6";
+            default: return "x?";
```
jordancarlin: If it reaches the default case, it should probably exit with an error instead of printing something that doesn’t make sense.

PR #1749, opened 2026-04-23
Add hypervisor CSR RVVI support and unit coverage

TnnsBeast:
Add H-extension and VS CSR export coverage to the RVVI paths used by lockstep, including mtinst/mtval2, H CSRs, VS CSRs, and RV32 high-half aliases. Expand the synthesizable RVVI CSR map and preserve the original case-table style in csrindextoaddr.

Fix hypervisor CSR behavior to better match the spec: legalize hedeleg according to IALIGN and read-only-zero bits, alias hie into mie and propagate mie writes into H interrupt-enable bits, handle explicit writes to vsepc and vstval correctly, legalize vstvec base alignment for direct and vectored modes, and use sepc for HS-mode sret while reserving vsepc for VS-mode sret. All of these changes are based on debugging with ImperasDV lockstep.

Update the testbench RVVI wiring so lockstep can compare the new CSR set, including correcting the mtinst CSR export slot.

Add hypervisorUnitTests coverage for implemented H/VS CSR read/write behavior, WARL masks, interrupt CSR aliasing, hgeip read-only behavior, HS-mode execution of hfence.gvma, and HFENCE decode legality. Add coverage README documentation with build/run notes for ImperasDV lockstep.


testbench/common/rvvitbwrapper.sv
```diff
@@ -114,7 +114,7 @@ module rvvitbwrapper import cvw::*; #(parameter cvw_t P,
   assign CSRArray[17] = 0; // 12'hF12
   assign CSRArray[18] = {{P.XLEN-12{1'b0}}, 12'h100}; //P.XLEN'h100; // 12'hF13
   assign CSRArray[19] = 0; // 12'hF15
-  assign CSRArray[20] = 0; // 12'h34A
+  assign CSRArray[20] = dut.core.priv.priv.csr.csrh.csrh.MTINST_REGW; // 12'h34A
```
jordancarlin: Does this need to be gated on `P.H_SUPPORTED`?

TnnsBeast: Yes, good catch. Updated

src/privileged/csrh.sv
```diff
@@ -444,7 +450,7 @@ module csrh import cvw::*;  #(parameter cvw_t P) (
       if (HIDELEG_REGW[10]) NextHIE[10] = CSRWriteValM[9]; // SEIE -> VSEIE
     end
   end
-  flopenr #(P.XLEN) HIEreg(clk, reset, (WriteHIEM | WriteVSIEM), NextHIE, HIE_REGW);
+  flopenr #(P.XLEN) HIEreg(clk, reset, (WriteHIEM | WriteHIEAliasM | WriteVSIEM), NextHIE, HIE_REGW);
```
davidharrishmc: No need for XLEN bits here.  I think all this state belongs in MIE, and then a subset is viewed from H, the sway a SIE does a subset.  There should be < 16 bits total.

src/privileged/csri.sv
```diff
@@ -81,9 +83,15 @@ module csri import cvw::*;  #(parameter cvw_t P) (
     else if (WriteMIPM) MIP_REGW_writeable <= (CSRWriteValM[11:0] & MIP_WRITE_MASK);
     else if (WriteSIPM) MIP_REGW_writeable <= (CSRWriteValM[11:0] & SIP_WRITE_MASK) | (MIP_REGW_writeable & ~SIP_WRITE_MASK);
   always_ff @(posedge clk)
-    if (reset)          MIE_REGW <= 12'b0;
-    else if (WriteMIEM) MIE_REGW <= (CSRWriteValM[11:0] & MIE_WRITE_MASK); // MIE controls M and S fields
-    else if (WriteSIEM) MIE_REGW <= (CSRWriteValM[11:0] & 12'h222 & MIDELEG_REGW) | (MIE_REGW & 12'h888); // only S fields
+    if (reset)          MIE_REGW_INT <= 12'b0;
+    else if (WriteMIEM) MIE_REGW_INT <= (CSRWriteValM[11:0] & MIE_WRITE_MASK); // MIE storage controls M/S fields; H fields alias hie
```
davidharrishmc: Seems this may need to be adjusted to handle additional H interrupts.

testbench/common/rvvitbwrapper.sv
```diff
@@ -133,6 +137,42 @@ module rvvitbwrapper import cvw::*; #(parameter cvw_t P,
...
+    assign CSRArray[57] = dut.core.priv.priv.csr.csrh.csrh.VSSCRATCH_REGW; // 12'h240
+    assign CSRArray[58] = dut.core.priv.priv.csr.csrh.csrh.VSEPC_REGW; // 12'h241
+    assign CSRArray[59] = dut.core.priv.priv.csr.csrh.csrh.VSCAUSE_REGW; // 12'h242
+    assign CSRArray[60] = dut.core.priv.priv.csr.csrh.csrh.VSTVAL_REGW; // 12'h243
+    assign CSRArray[61] = dut.core.priv.priv.csr.csrh.csrh.VSIP_REGW; // 12'h244
+    assign CSRArray[62] = dut.core.priv.priv.csr.csrh.csrh.VSATP_REGW; // 12'h280
+    assign CSRArray[63] = dut.core.priv.priv.csr.csrh.csrh.VSTIMECMP_REGW[P.XLEN-1:0]; // 12'h24D
+    assign CSRArray[64] = (P.XLEN == 32) ? dut.core.priv.priv.csr.csrh.csrh.VSTIMECMP_REGW[63:32] : '0; // 12'h25D
+  end else begin : no_hypervisor_csrs
+    for (genvar csrIdx = 36; csrIdx < TOTAL_CSRS; csrIdx++) begin
```
davidharrishmc: Change to 65, so if TOTAL_CSRS changes later for another extension, this doesn't break.

src/privileged/csri.sv
```diff
@@ -71,27 +72,54 @@ module csri import cvw::*;  #(parameter cvw_t P) (
       assign STIP = MIP_REGW_writeable[5];
     end
     assign SIP_WRITE_MASK = 12'h002 & MIDELEG_REGW; // SSIP is writeable in SIP (privileged 20210108-draft 4.1.3)
-    assign MIE_WRITE_MASK = 12'hAAA;
+    assign MIE_WRITE_MASK = 12'hAAA | HIE_WRITE_MASK;
   end else begin:mask
     assign MIP_WRITE_MASK = 12'h000;
     assign SIP_WRITE_MASK = 12'h000;
-    assign MIE_WRITE_MASK = 12'h888;
+    assign MIE_WRITE_MASK = 12'h888 | HIE_WRITE_MASK;
```
davidharrishmc: If S is not supported, then HIE_WRITE_MASK should be 0 and this is unnecessary.

src/privileged/csri.sv
```diff
@@ -35,17 +35,18 @@ module csri import cvw::*;  #(parameter cvw_t P) (
...
   input  logic [11:0]       MIDELEG_REGW,
   input  logic              ENVCFG_STCE,
+  output logic [11:0]       HIE_REGW,
   output logic [11:0]       MIP_REGW, MIE_REGW,
   output logic [11:0]       MIP_REGW_writeable // only SEIP, STIP, SSIP are actually writeable; the rest are hardwired to 0
 );
 
-  logic [11:0]              MIP_WRITE_MASK, SIP_WRITE_MASK, MIE_WRITE_MASK;
-  logic                     WriteMIPM, WriteMIEM, WriteSIPM, WriteSIEM;
+  logic [11:0]              MIP_WRITE_MASK, SIP_WRITE_MASK, MIE_WRITE_MASK, SIE_WRITE_MASK, HIE_WRITE_MASK, NextMIE_REGW;
```
davidharrishmc: SGEIE is bit 12.  How does this all fit in 12 bits?  Let's bump the register up to 16 bits because we will need LCOFIE in bit 13 eventually anyway.

davidharrishmc (changes requested):

PR #1755, opened 2026-05-02
Support for Opensuse 16

fourth-bit:
Adds support for Opensuse leap 16. We tested it by installing the toolchain in a podman container. 

Coauthored With @Lamarr3314

### DIFF
--- a/bin/wally-environment-check.sh
+++ b/bin/wally-environment-check.sh
@@ -132,11 +132,11 @@ elif [[ "$ID" == opensuse-leap || "$ID" == sles || "$ID_LIKE" == *suse* ]]; then
     fi
     export SUSE_VERSION="${VERSION_ID//.}"
     if (( SUSE_VERSION < 156 )); then
-        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
+        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE versions 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
         exit 1
     fi
     if (( SUSE_VERSION > 160 )); then
-        printf "${WARNING_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have a newer version ($VERSION). The installation for SUSE 15.6 will be attempted, but there may be issues."
+        printf "${WARNING_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE versions 15.6 and 16.0. You have a newer version ($VERSION). The installation for SUSE 16.0 will be attempted, but there may be issues."
     fi
 else
     printf "${FAIL_COLOR}%s%s%s\n${ENDC}" "The Wally installation script is currently only compatible with Ubuntu, Debian, SUSE, and Red Hat family " \
--- a/bin/wally-environment-check.sh
+++ b/bin/wally-environment-check.sh
@@ -132,11 +132,11 @@ elif [[ "$ID" == opensuse-leap || "$ID" == sles || "$ID_LIKE" == *suse* ]]; then
     fi
     export SUSE_VERSION="${VERSION_ID//.}"
     if (( SUSE_VERSION < 156 )); then
-        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6. You have version $VERSION. Please upgrade to a supported version of SUSE."
+        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
         exit 1
     fi
-    if (( SUSE_VERSION > 156 )); then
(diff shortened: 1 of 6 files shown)


bin/wally-environment-check.sh
```diff
@@ -132,11 +132,11 @@ elif [[ "$ID" == opensuse-leap || "$ID" == sles || "$ID_LIKE" == *suse* ]]; then
...
     export SUSE_VERSION="${VERSION_ID//.}"
     if (( SUSE_VERSION < 156 )); then
-        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6. You have version $VERSION. Please upgrade to a supported version of SUSE."
+        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
         exit 1
     fi
-    if (( SUSE_VERSION > 156 )); then
-        printf "${WARNING_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6. You have a newer version ($VERSION). The installation for SUSE 15.6 will be attempted, but there may be issues."
+    if (( SUSE_VERSION > 160 )); then
+        printf "${WARNING_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have a newer version ($VERSION). The installation for SUSE 15.6 will be attempted, but there may be issues."
```
jordancarlin:
```suggestion
        printf "${WARNING_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE versions 15.6 and 16.0. You have a newer version ($VERSION). The installation for SUSE 15.6 will be attempted, but there may be issues."
```

bin/wally-environment-check.sh
```diff
@@ -132,11 +132,11 @@ elif [[ "$ID" == opensuse-leap || "$ID" == sles || "$ID_LIKE" == *suse* ]]; then
     fi
     export SUSE_VERSION="${VERSION_ID//.}"
     if (( SUSE_VERSION < 156 )); then
-        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6. You have version $VERSION. Please upgrade to a supported version of SUSE."
+        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE version 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
```
jordancarlin:
```suggestion
        printf "${FAIL_COLOR}%s\n${ENDC}" "The Wally installation script has only been tested with SUSE versions 15.6 and 16.0. You have version $VERSION. Please upgrade to a supported version of SUSE."
```

bin/wally-package-install.sh
```diff
@@ -102,18 +102,33 @@ case "$FAMILY" in
...
+            PYTHON_VERSION_PACKAGE=python313
+        elif ((SUSE_VERSION >= 156)); then
+            PYTHON_VERSION=python3.12
+            PYTHON_VERSION_PACKAGE=python312
+        fi
+
         PACKAGE_MANAGER="zypper -n"
         UPDATE_COMMAND="$PACKAGE_MANAGER update"
         GENERAL_PACKAGES+=(which curl "$PYTHON_VERSION_PACKAGE" "$PYTHON_VERSION_PACKAGE"-pip pkg-config)
+        GENERAL_PACKAGES+=(gcc-c++)
```
jordancarlin: Combine these lists

jordancarlin (changes requested):
A few minor nits, but otherwise this looks great!

jordancarlin (approved):
LGTM once CI passes.

jordancarlin:
@fourth-bit this seems to have broken OpenSUSE 15.6. See https://github.com/openhwgroup/cvw/actions/runs/25287064089/job/74133151567?pr=1755.

fourth-bit:
I just checked the installation logs from the main branch here: https://github.com/openhwgroup/cvw/actions/runs/25096807571/job/73535533403, and this error was appearing before this PR. It looks like there was a breaking change to the `ncurses-devel` and `ncurses5-devel` package in the Opensuse 15.6 repos recently. Right now, we cannot install `ncurses-devel` or `ncurses5-devel` on Opensuse 15.6 without getting the error that we see. Passing `--force-resolution` or `--allow-downgrade` seems to get it to install all of the packages. Should I add a fix for opensuse 15.6 in this PR, or open a new one addressing it?

jordancarlin:
`--force-resolution` seems like a flag we should try to avoid unless there is a clear reason it is needed. Can you look around online to see if this is as known issue with ncurses? Do we need to be installing a different version of it or something?

fourth-bit:
I looked through the opensuse bug reporting and did some general searches and did not find a reference to this bug online. The problem is that the container that we use comes with libnucurses6 built from ncurses-6.1-150000.5.33.1.src and is version 6.1-150000.5.33.1, but the ncurses-devel package has a dependency of libncurses6 = 6.1-150000.5.30.1 . An older version of these packages will have the same issue because the container comes with this version of libncurses that is too new. `ncurses-devel` is a dependency of `ncurses5-devel`, so going back to these older packages also isn't possible.

fourth-bit:
The newest version of the opensuse leap image, 15.6.7.73 has this issue while 15.6.7.64 does not

jordancarlin:
Sounds like an issue with the distro itself then. Give it a few weeks and then check back in to see if we need to change anything to get it working.

PR #1761, opened 2026-05-10
Implement bare mode HLV/HLVX/HSV path and coverage

TnnsBeast:
- Enables basic hypervisor load/store instruction support for the implemented Bare/Bare translation case
- Adds ImperasDV lockstep coverage tests for the main HLV/HLVX/HSV load and store sizes, sign-extension behavior, and U-mode access through hstatus.HU.

### DIFF
--- a/src/ieu/controller.sv
+++ b/src/ieu/controller.sv
@@ -296,11 +296,12 @@ module controller import cvw::*;  #(parameter cvw_t P) (
                       ControlsD = `CTRLW'b1_000_01_00_000_0_0_1_1_0_0_0_0_0_00_0_0; // jalr
       7'b1101111:     ControlsD = `CTRLW'b1_011_11_00_000_0_0_1_1_0_0_0_0_0_00_0_0; // jal
       7'b1110011: if (P.ZICSR_SUPPORTED) begin
-                   if (HLVHSVInstrD & HLVHSVLoadD)
-                      ControlsD = `CTRLW'b1_101_01_10_001_0_0_0_0_0_0_0_0_0_00_0_0; // HLV/HLVX: zero-offset virtual-machine load
-                   else if (HLVHSVInstrD)
-                      ControlsD = `CTRLW'b0_101_01_01_000_0_0_0_0_0_0_0_0_0_00_0_0; // HSV: zero-offset virtual-machine store
-                   else if (PFunctD)
+                   if (HLVHSVInstrD) begin
+                     if (HLVHSVLoadD)
+                       ControlsD = `CTRLW'b1_101_01_10_001_0_0_0_0_0_0_0_0_0_00_0_0; // HLV/HLVX: zero-offset virtual-machine load
+                     else
+                       ControlsD = `CTRLW'b0_101_01_01_000_0_0_0_0_0_0_0_0_0_00_0_0; // HSV: zero-offset virtual-machine store
+                   end else if (PFunctD)
                       ControlsD = `CTRLW'b0_000_00_00_000_0_0_0_0_0_0_1_0_0_00_0_0; // privileged; decoded further in privdec modules
                    else if (CSRFunctD)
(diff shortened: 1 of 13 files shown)


src/ieu/controller.sv
```diff
@@ -282,7 +296,11 @@ module controller import cvw::*;  #(parameter cvw_t P) (
                       ControlsD = `CTRLW'b1_000_01_00_000_0_0_1_1_0_0_0_0_0_00_0_0; // jalr
       7'b1101111:     ControlsD = `CTRLW'b1_011_11_00_000_0_0_1_1_0_0_0_0_0_00_0_0; // jal
       7'b1110011: if (P.ZICSR_SUPPORTED) begin
-                   if (PFunctD)
+                   if (HLVHSVInstrD & HLVHSVLoadD)
```
davidharrishmc: This would be cleaner to have a if (HLSVHSVInstrD) top level, with if (HSLHSVLoadD) and else inside.

src/ieu/extend.sv
```diff
@@ -47,8 +47,8 @@ module extend import cvw::*;  #(parameter cvw_t P) (
       3'b011:   ImmExtD = {{(P.XLEN-20){InstrD[31]}}, InstrD[19:12], InstrD[20], InstrD[30:21], 1'b0};
       // U-type (lui, auipc)
       3'b100:   ImmExtD = {{(P.XLEN-31){InstrD[31]}}, InstrD[30:12], 12'b0};
-      // Store Conditional: zero offset
-      3'b101:  if (P.ZALRSC_SUPPORTED | P.ZAAMO_SUPPORTED | P.ZICBOM_SUPPORTED | P.ZICBOZ_SUPPORTED) ImmExtD = '0;
+      // Store Conditional, AMO, CMO, HLV/HLVX/HSV: zero offset
```
davidharrishmc: AMO -> Atomics

src/ifu/ifu.sv
```diff
@@ -187,7 +187,8 @@ module ifu import cvw::*;  #(parameter cvw_t P) (
     assign TLBFlush = sfencevmaM & ~StallMQ;
 
     mmu #(.P(P), .TLB_ENTRIES(P.ITLB_ENTRIES), .IMMU(1))
-    immu(.clk, .reset, .SATP_REGW, .STATUS_MXR, .STATUS_SUM, .STATUS_MPRV, .STATUS_MPP, .ENVCFG_PBMTE, .ENVCFG_ADUE,
+    immu(.clk, .reset, .SATP_REGW, .STATUS_MXR, .STATUS_SUM, .STATUS_MPRV, .STATUS_MPP,
+         .HSTATUS_SPVP(1'b0), .HLVHSVLegalM(1'b0), .ENVCFG_PBMTE, .ENVCFG_ADUE,
```
davidharrishmc, line 193: Is it hard to hook up SPVP and HLVHSVLegal?

davidharrishmc (changes requested):

src/privileged/csrhvirt.sv
davidharrishmc: Hypervisor is not documented in RISC-V SoCD

davidharrishmc, line 2: What is the purpose of csrh vs. csrhvirt?  Would it be cleaner for them to both be in csrh?

davidharrishmc:
Looking good!

src/ifu/ifu.sv
TnnsBeast: These are intentionally tied off for now since IMMU doesn't use them, but full VS/VU instruction fetch would need separate two-stage translation plumbing (added comment in recent commit to clarify this).

src/privileged/csrhvirt.sv
TnnsBeast: I think it is cleaner to keep them separate to avoid bloating csrh (which was a piece of feedback from a few PRs ago). csrhvirt contains combinational logic for CSR address remapping, virtual instruction classification, and trap GVA generation that partitions cleanly from csrh.

davidharrishmc (approved):

PR #1763, opened 2026-05-15
Add support for Ubuntu 26.04

jordancarlin:
Did this a while ago but apparently forgot to open a PR.

Worked on with @fourth-bit and @Lamarr3314


jordancarlin:
@davidharrishmc this should be merged. The OpenSUSE error is unrelated.

Issue #1538, opened 2025-10-18
Found some bugs while running regression for SV48


Zain2050:
Addresses the problem faced in PR #1532. @davidharrishmc @rosethompson 

The test `sv48_res_global_pte_U_mode.S` normally passes but fails when `sv48_pte_reserved_field_S_mode.S` is executed prior to it. This seemed really strange. I looked into it and have found some bugs.

It seems like BTB is the root cause of the problem. 

**Normal Execution:**
<img width="423" height="113" alt="Image" src="https://github.com/user-attachments/assets/cd8faffc-b53f-43ee-acac-d5d84ccbcdc2" />

**If sv48_pte_reserved_field_S_mode.S is executed prior to it:**

<img width="403" height="112" alt="Image" src="https://github.com/user-attachments/assets/e796d7bc-e6cd-4495-898c-e1137b7d5574" />

Initially, I observed that there was a branch misprediction, but the instruction in fetch stage was lw. Therefore, seemed strange. I then checked the other test (pte_reserved_field) and it has a jalr at that address.

```
[1603] [S]: 0x0000030080000EC0 (0x000780E7) jalr x1, 0x0(x15)
x1 <- 0x0000030080000EC4
```
This is not being cleared from the BTB and it's messing with the other test. This is causing BPJumpF to go high and jump to an address from the BTB despite having lw instead of a control flow instruction in fetch stage.
Wally wasn't supposed to make this prediction but it eventually recovers and jumps back to the correct address. But due to this bug I have encountered 2 other bugs which were the reason for signature mismatch.

When we get the (buggy) misprediction due to lw in fetch stage, there is a sw in memory stage. The sw causes a DTLBmiss while at the same time we get an ITLBmiss as well. The HPTW first calculates physical address for sw and as soon as the walk finishes and 0x40 (desired value) is going to be stored in DCache, the ITLBmiss causes a HPTW flush which in turn flushes LSU.

`  assign HPTWFlushW = (WalkerState == IDLE & TLBMissOrUpdateDA) | (WalkerState != IDLE & HPTWFaultM);`

Due to LSUFlushW, CacheEn goes low and sw is unable to write the desired value to DCache. 

<img width="1266" height="142" alt="Image" src="https://github.com/user-attachments/assets/36d3deb9-4c7a-4288-b92e-be2d3a58be39" />

[...]

PR #1765, opened 2026-05-16
catch all reserved bits in TLB

davidharrishmc:


src/mmu/tlb/tlbram.sv
```diff
@@ -51,6 +51,9 @@ module tlbram import cvw::*;  #(parameter cvw_t P,
   or_rows #(TLB_ENTRIES, P.XLEN) PTEOr(RamRead, PageTableEntry);
 
   // Rename the bits read from the TLB RAM
-  assign PTEAccessBits = {PageTableEntry[P.XLEN-1:P.XLEN-4] & {4{P.XLEN == 64}}, PageTableEntry[7:0]}; // for RV64 include N and PBMT bits and OR of reserved bits
+  // PTEAccessBits[11:8] = {N, PBMT[1:0], reserved}
+  // reserved = OR of PTE[60:54]; catches all reserved bits for SV39/SV48/SV57
+  assign PTEAccessBits = {(P.XLEN == 64 ?
+    {PageTableEntry[P.XLEN-1:P.XLEN-3], |PageTableEntry[P.XLEN-4:P.XLEN-10]} : 4'b0), PageTableEntry[7:0]};
```
jordancarlin, line 57: If Svnapot and/or Svpbmt are not enabled, should the bits corresponding to those extensions be included in the reserved bits that are ORed together?

jordancarlin (commented):
How was this found/what prompted this?

src/mmu/tlb/tlbram.sv
davidharrishmc: These other bits are already checked explicitly in tlbcontrol.sv, including whether they are supported. I'd rather not make bigger changes to code that already works.

davidharrishmc: This loss of checking reserved bits was identified by AI while troubleshooting #1538.  It is not the root cause, but might as well be fixed.

davidharrishmc:
@rosethompson Jordan has graduated and is on travel.  Can you review and merge?

jordancarlin (approved):
Approving, but I am concerned why this does not cause the ACT4 Sv tests to fail. I thought those had tests for the reserved bits. Either we need to expand those tests (maybe we are only testing some of the reserved bits?) or this change isn't actually needed and we catch the reserved bits elsewhere in Wally.

davidharrishmc:
Good point.  @Zain2050 do you know if we are testing all of the RV64 reserved PTE bits individually?

davidharrishmc:
Turns out that cvw tlbramline.sv does check all the reserved bits, and so does sv39_pte_reserved_field_Smode.S.
https://github.com/riscv/riscv-arch-test/issues/1559

Closing this PR.

PR #1767, opened 2026-05-20
Bump qemu to 11.0.1

fourth-bit:


fourth-bit:
The ubuntu-arm error looks like something weird happened in CI that is unrelated to the changes. The opensuse 15.6 install failure has been reported to opensuse here: https://bugzilla.opensuse.org/show_bug.cgi?id=1266957

PR #1768, opened 2026-05-22
mmu: preserve global TLB entries on ASID sfence.vma; harden HPTW vs ITLB-miss store drop (#1538)

### LINKED ISSUE #1538: Found some bugs while running regression for SV48

davidharrishmc:
## Fix #1538: sv48 `res_global` PTE U-mode test fails after cross-test residue

  Closes #1538.

  ### Background

  `sv48_res_global_pte_U_mode.S` passes in isolation but fails when `sv48_pte_reserved_field_S_mode.S` runs first in the 
  same simulation (reads stale `0xffffffffbeefcaf2` at signature `0x80009150`, expected `0x40`). Both tests target 
  `va_data+20 → PA=0x80000414`. The failure exposes three independent bugs across the MMU and D-cache, all of which can in
  principle be hit by any workload that combines an ASID-only `sfence.vma`, an HPTW walk, and a closely-following load/store
   at a different set.

  Two reproducer suites are wired in `testbench/tests.vh`: `arch64vm_sv48_a` (passing order) and `arch64vm_sv48_b` (failing
  order).

  ### Three independent fixes

  #### 1. `1867e1cce` — MMU: preserve global TLB entries on ASID `sfence.vma`; defer ITLB walk during in-flight data op

  - `sfence.vma rs1, rs2` previously flushed every TLB entry unconditionally. The new `sfencevmaAllM` signal is asserted
  only when both rs1 and rs2 are `x0`, so the standard ASID-only sfence keeps PTEs with `PTE_G=1` — matching the priv spec.
  - When a data-stage operation is mid-flight on the bus, the HPTW's ITLB-miss walk is deferred until `DCacheBusStallM`
  clears. Without this, an ITLB walk could stomp on an in-flight store and silently drop it.

  #### 2. `d95629605` — D$: stalled store-hit data loss + faulting-store commit

  Two D$ hazards:

  - **Stalled store-hit loss.** A new signal `StoreHitFirstStall = CacheRW[0] & Stall & ~WasStalling & ~StallConditions`
  extends `CacheEn` for the first stall cycle of a store-hit so the SRAM word write actually fires before the stall blocks
  the FSM. `WasStalling` (flopped `Stall`) prevents re-firing on subsequent stall cycles. `SelAdrTag` is narrowed to the
  first stall cycle so the tag-SRAM pre-fetch is not blocked on later cycles.
  - **Faulting store committed to cache.** The `CacheEn` extension is gated by `~StoreAmoPageFaultM & ~FlushStage` so a
  store that takes a page fault never updates the cache. `StoreAmoPageFaultM` is threaded `lsu → cache → cachefsm`; `ifu.sv`
   ties it `1'b0` for the I$.

  Files touched: `src/cache/cachefsm.sv`, `src/cache/cache.sv`, `src/lsu/lsu.sv`, `src/ifu/ifu.sv`.

  #### 3. (new) — D$: detect stale tag/data SRAM read after HPTW completes

  Tag and data SRAM reads are flopped: in any given cycle `Hit` reflects the SRAM read launched in the previous cycle,
  indexed by the previous `CacheSetTag`. When the HPTW finishes a walk and the LSU re-issues its access on the next cycle,
  the SRAM still returns the HPTW's set; if the LSU's new `PAdr` maps to a different set, `Hit` is computed against the
  wrong tag — a false miss fills clean DRAM data over the LSU's existing dirty line and silently loses the store. This is
  what produces the iter-4 dirty-line loss in `arch64vm_sv48_b` immediately after `sfence.vma x0, ASID`.

  Fix in `src/cache/cache.sv` + `src/cache/cachefsm.sv`:

  - Flop `CacheSetTag` → `CacheSetTagPrev` and `SelHPTW` → `SelHPTWPrev` (gated by `CacheEn`).
  - `TagSetStale = SelHPTWPrev & ~SelHPTW & (CacheSetTagPrev != PAdr_set)`.
  - On the stale cycle: suppress `AnyMiss` / `AnyUpdateHit` / `AnyHit`, add to `StallConditions` (one-cycle stall), and add
  to `SelAdrTag` + `SelAdrData` so both SRAMs re-read at `PAdr`'s set. Next cycle `Hit` is valid.

  The narrow `SelHPTWPrev & ~SelHPTW` predicate is required: a broader form (just the set-mismatch check) regresses
  `arch64m`, because during MDU stalls the E-stage and M-stage hold different instructions, so `CacheSetTagPrev` (captured
  from `NextSet` = E-stage address) legitimately differs from `PAdr` (= M-stage address) every cycle. Restricting to the
  HPTW→LSU transition makes the stall fire only when there's actually a stale read.

  I$ has `SelHPTW = '0` hardwired (`src/ifu/ifu.sv:260`), so `SelHPTWPrev` is always 0 and the new logic is dead for the
  I-cache — no I$ behavior change.

  ### Test plan

  Verilator on `rv64gc`:

  - [x] `arch64vm_sv48_a` (original passing order)
  - [x] `arch64vm_sv48_b` (failing order — the #1538 reproducer)
  - [x] `arch64vm_sv48` (full), `arch64vm_sv48_mxr_isolate`
  - [x] `arch64vm_sv39`, `arch64vm_sv39_isolate`, `arch64vm_sv57`
  - [x] `arch64i`, `arch64m`, `arch64c`, `arch64a_amo`, `arch64f`, `arch64d`
  - [x] `arch64priv`, `arch64pmp`, `arch64zicboz`, `arch64zifencei`
  - [x] `arch64zba`, `arch64zbb`, `arch64zbs`, `arch64zcb`
  - [x] `wally64priv`

  Verilator on `rv32gc`:

  - [x] `arch32i`, `arch32priv`, `wally32priv`

  Not yet re-run (recommended for nightly): `buildroot` Linux boot, embench. The new TagSetStale logic only fires on
  HPTW→LSU set-changing transitions; expected harmless but worth a long-workload pass.

### DIFF
--- a/sim/questa/coverage-exclusions-rv64gc.do
+++ b/sim/questa/coverage-exclusions-rv64gc.do
@@ -90,7 +90,7 @@ coverage exclude -scope /dut/core/ifu/bus/icache/icache/cachefsm -ftrans CurrSta
 # exclude unused transitions from case statement. Unfortunately the whole branch needs to be excluded I think. Expression coverage should still work.
 coverage exclude -scope /dut/core/ifu/bus/icache/icache/cachefsm -linerange [GetLineNum ${SRC}/cache/cachefsm.sv "exclusion-tag: icache state-case"] -item b 1
 # I$ does not flush
-coverage exclude -scope /dut/core/ifu/bus/icache/icache/cachefsm -linerange [GetLineNum ${SRC}/cache/cachefsm.sv "exclusion-tag: icache FlushCache"] -item e 1 -fecexprrow 2
+coverage exclude -scope /dut/core/ifu/bus/icache/icache/cachefsm -linerange [GetLineNum ${SRC}/cache/cachefsm.sv "exclusion-tag: icache FlushCache"] -item e 1 -fecexprrow 2 6
 coverage exclude -scope /dut/core/ifu/bus/icache/icache/AdrSelMuxLRU -item b 1
 # exclude branch/condition coverage: LineDirty if statement
 coverage exclude -scope /dut/core/ifu/bus/icache/icache/cachefsm -linerange [GetLineNum ${SRC}/cache/cachefsm.sv "exclusion-tag: icache FETCHStatement"] -item bc 1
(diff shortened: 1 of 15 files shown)


rosethompson:
I will accept the merge, but we need to address the changes to the cache.  They introduce extra logic that should not be necessary.

Issue #1780, opened 2026-06-09
WFI illegal-instruction trap on the next instruction when delegated to S-mode


Teemooooooooo:
When running in U mode, WFI is expected to trigger illegal instruction exception when exceeding the timeout limit. CVW shows mismatch with imperas on which instruction is trapped at when illegal instruction is delegated to S mode. CVW incorrectly shows that the trap happens at the nop after wfi when it was supposed to trap on wfi. This problem goes away when medeleg is cleared before the test (all traps handled at M mode).

The command to reproduce the bug: 

wsim --sim verilator --sim questa --lockstepverbose --args '+UART_LOG=1 +UART_LOG_FILE=InterruptsU-00.summary.log' rv64gc --elf InterruptsU-00.elf

The elf is attached below: 

[InterruptsU-00.elf.gz](https://github.com/user-attachments/files/28763809/InterruptsU-00.elf.gz)

PR #1783, opened 2026-06-09
Fix: WFI illegal-instruction trap on the next instruction when delegated to S-mode

Teemooooooooo:
Fix to [issue 1780](https://github.com/openhwgroup/cvw/issues/1780)

Root cause to the issue: WFITimeoutM remains high for another cycle after it hits the timeout limit. 

Added in logic to ensure that WFITimeoutM clears after the timeout limit is hit :)

### DIFF
--- a/src/privileged/privdec.sv
+++ b/src/privileged/privdec.sv
@@ -88,11 +88,11 @@ module privdec import cvw::*;  #(parameter cvw_t P) (
 
   if (P.U_SUPPORTED) begin : wfi
     logic [P.WFI_TIMEOUT_BIT:0] WFICount, WFICountPlus1;
-    assign WFICountPlus1 = (wfiM & (~WFICount[P.WFI_TIMEOUT_BIT])) ? WFICount + 1 : '0; // restart counting on WFI
+    assign WFICountPlus1 = wfiM ? WFICount + 1 : '0; // restart counting on WFI
     flopr #(P.WFI_TIMEOUT_BIT+1) wficountreg(clk, reset, WFICountPlus1, WFICount);  // count while in WFI
   // coverage off -item e 1 -fecexprrow 1
   // WFI Timeout trap will not occur when STATUS_TW is low while in supervisor mode, so the system gets stuck waiting for an interrupt and triggers a watchdog timeout.
-    assign WFITimeoutM = ((STATUS_TW & PrivilegeModeW != P.M_MODE) | (P.S_SUPPORTED & PrivilegeModeW == P.U_MODE)) & WFICount[P.WFI_TIMEOUT_BIT];
+    assign WFITimeoutM = wfiM & (((STATUS_TW & PrivilegeModeW != P.M_MODE) | (P.S_SUPPORTED & PrivilegeModeW == P.U_MODE)) & WFICount[P.WFI_TIMEOUT_BIT]);
   // coverage on
   end else assign WFITimeoutM = 1'b0;
 
--- a/src/privileged/privdec.sv
+++ b/src/privileged/privdec.sv
@@ -88,7 +88,7 @@ module privdec import cvw::*;  #(parameter cvw_t P) (
 
   if (P.U_SUPPORTED) begin : wfi
     logic [P.WFI_TIMEOUT_BIT:0] WFICount, WFICountPlus1;
-    assign WFICountPlus1 = wfiM ? WFICount + 1 : '0; // restart counting on WFI
+    assign WFICountPlus1 = (wfiM & (~WFICount[P.WFI_TIMEOUT_BIT])) ? WFICount + 1 : '0; // restart counting on WFI
     flopr #(P.WFI_TIMEOUT_BIT+1) wficountreg(clk, reset, WFICountPlus1, WFICount);  // count while in WFI
   // coverage off -item e 1 -fecexprrow 1
   // WFI Timeout trap will not occur when STATUS_TW is low while in supervisor mode, so the system gets stuck waiting for an interrupt and triggers a watchdog timeout.
--- a/src/privileged/privdec.sv
+++ b/src/privileged/privdec.sv
@@ -37,6 +37,7 @@ module privdec import cvw::*;  #(parameter cvw_t P) (
   input  logic         IllegalCSRAccessM,                   // Not a legal CSR access
   input  logic [1:0]   PrivilegeModeW,                      // current privilege level
   input  logic         STATUS_TSR, STATUS_TVM, STATUS_TW,   // status bits
+  input  logic         TrapM,                               // Trap is occurring
   output logic         IllegalInstrFaultM,                  // Illegal instruction
   output logic         EcallFaultM, BreakpointFaultM,       // Ecall or breakpoint; must retire, so don't flush it when the trap occurs
   output logic         sretM, mretM, RetM,                  // return instructions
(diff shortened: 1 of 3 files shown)


davidharrishmc:
@Teemooooooooo great to find such a simpler fix.  Claude thinks there is an even more robust fix.  Could you review its feedback critically and implement if you agree?

❯ Evaluate the fix in https://github.com/openhwgroup/cvw/pull/1783 to the issue in 1780.  Explain why it works, why the   
  issue was only seen when illegal instructions are delegated, and whether the fix might be hiding other issues.  Give me 
  critical feedback; do not presume the fix is right.                                                                     
  

● Verdict

  The fix is functionally correct for the reported failure, and the PR author's root-cause statement ("WFITimeoutM remains
  high for another cycle") is accurate but shallow. The deeper defect is that WFITimeoutM is an exception signal derived
  from persistent counter state plus the current privilege mode, with no qualification that a WFI (or any valid
  instruction) is actually in the M stage. The PR makes the stale state disappear one cycle sooner rather than fixing that
  qualification, and in doing so it converts the timeout signal into a single-cycle pulse, which introduces a (remote)
  new corner case. I'd ask for a revision.

  The actual failure mechanism, cycle by cycle

  Call cycle T the cycle WFICount[16] first sets, with the WFI stalled in M (WFIStallM, hazard.sv:49) and the following
  nop stalled in E.

  Cycle T (correct trap): WFITimeoutM → IllegalInstrFaultM (privdec.sv:113) → TrapM. The trap is taken properly: sepc ←
  PCM = the WFI's address, delegation sends it to S-mode, the pipeline flushes. But two pieces of state update badly at
  this same clock edge:

  1. wfiM is still high during T, so the old logic WFICountPlus1 = wfiM ? WFICount + 1 : '0 loads WFICount with timeout+1
  — bit 16 stays set into T+1. The counter only clears at the end of T+1, one cycle after the wfi is flushed.
  2. PCMReg (ifu.sv:420) is a flopenr with no flush input. TrapM releases StallM, so PCM ← PCE = the address of the
  flushed nop.

  Cycle T+1 (spurious trap): The M stage holds a bubble, but
  WFITimeoutM = ((STATUS_TW & PrivilegeModeW != M_MODE) | (S_SUPPORTED & PrivilegeModeW == U_MODE)) & WFICount[16]
  (privdec.sv:95) is gated by neither wfiM nor InstrValidM — and neither is ExceptionM in trap.sv:87. With the stale count
  bit still set, the privilege condition decides everything. After delegation, PrivilegeModeW = S_MODE and the test holds
  mstatus.TW = 1 (the cp_wfi_timeout block in InterruptsU-00.S explicitly fixes TW=1), so the first term is true: a
  second illegal-instruction trap fires on the pipeline bubble, writing sepc ← PCM = the nop's address and re-vectoring to
  stvec. That is exactly the "trap on the nop after wfi" the lockstep flagged.

  Why only when delegated

  The stale counter bit exists in both cases. What differs is the privilege mode during the stale cycle T+1:

  - Not delegated: PrivilegeModeW = M_MODE in T+1. Both terms of the condition are false (TW & priv≠M fails, priv==U
  fails), so the stale bit is masked. The bug is present but invisible.
  - Delegated: PrivilegeModeW = S_MODE in T+1, and with TW=1 the condition re-arms.

  So clearing medeleg doesn't avoid the bug — it just routes the machine into the one privilege mode that happens to mask
  it. Note also it needs TW=1 specifically: a pure U-mode timeout with TW=0 (trapping via the priv==U term) would also
  self-mask after entering S-mode, which is why this survived until a test crossed TW=1 with delegation.

  Why the fix works

  WFICountPlus1 = (wfiM & ~WFICount[16]) ? WFICount + 1 : '0 stops incrementing once bit 16 sets, so the counter resets at
  the end of T and bit 16 is high for exactly one cycle — the cycle the legitimate trap fires with the WFI genuinely in
  M. The stale T+1 cycle is gone. Timeout latency is unchanged (the bit still sets at the same count), and behavior in
  non-trapping modes (M-mode WFI, S-mode with TW=0, where WFI waits indefinitely for an interrupt) is unchanged — the
  counter saturate-resets instead of free-wrapping, both invisible. It also covers the race where an interrupt arrives the
  same cycle as the timeout.

  Critical feedback

  1. It patches the symptom, not the invariant. The violated invariant is: an M-stage exception must only assert when
  attributable to a valid instruction in M. Every other input to ExceptionM is derived from the M-stage instruction and
  dies with the flush; WFITimeoutM is the lone violator, and the PR leaves it un-gated — it just arranges for the stale
  state to vanish in time. The structurally honest fix is one literal:

  assign WFITimeoutM = wfiM & ((STATUS_TW & PrivilegeModeW != P.M_MODE) | ...) & WFICount[P.WFI_TIMEOUT_BIT];

  No WFI in M → no WFI timeout fault, regardless of counter or privilege history.

  2. The pulse introduces a trap-swallowing corner. TrapM = (ExceptionM & ~CommittedF) | InterruptM (trap.sv:93).
  Why the fix works

  WFICountPlus1 = (wfiM & ~WFICount[16]) ? WFICount + 1 : '0 stops incrementing once bit 16 sets, so the counter resets at
  the end of T and bit 16 is high for exactly one cycle — the cycle the legitimate trap fires with the WFI genuinely in
  M. The stale T+1 cycle is gone. Timeout latency is unchanged (the bit still sets at the same count), and behavior in
  non-trapping modes (M-mode WFI, S-mode with TW=0, where WFI waits indefinitely for an interrupt) is unchanged — the
  counter saturate-resets instead of free-wrapping, both invisible. It also covers the race where an interrupt arrives the
  same cycle as the timeout.

  Critical feedback

  1. It patches the symptom, not the invariant. The violated invariant is: an M-stage exception must only assert when
  attributable to a valid instruction in M. Every other input to ExceptionM is derived from the M-stage instruction and
  dies with the flush; WFITimeoutM is the lone violator, and the PR leaves it un-gated — it just arranges for the stale
  state to vanish in time. The structurally honest fix is one literal:

  assign WFITimeoutM = wfiM & ((STATUS_TW & PrivilegeModeW != P.M_MODE) | ...) & WFICount[P.WFI_TIMEOUT_BIT];

  No WFI in M → no WFI timeout fault, regardless of counter or privilege history.

  2. The pulse introduces a trap-swallowing corner. TrapM = (ExceptionM & ~CommittedF) | InterruptM (trap.sv:93).
  Post-fix, WFITimeoutM is high for exactly one cycle; if CommittedF happened to be high in that cycle, the trap is
  suppressed, the counter has already self-cleared, and the WFI silently waits another full 2^16 cycles. Pre-fix, the bit
  stayed set for 65k cycles so a suppressed trap fired as soon as CommittedF dropped. In practice an in-flight fetch
  completes early in a 65k-cycle stall, so this is remote — but the wfiM-gated version (keeping the counter saturated at
  the threshold rather than reset: wfiM ? (WFICount[16] ? WFICount : WFICount+1) : '0) makes the signal a level that holds
  until the trap actually retires the WFI, and has no such hazard.

  3. What it leaves hidden. Two latent issues this bug exposed remain after the merge: (a) ExceptionM in trap.sv is not
  qualified by InstrValidM — it currently works only because every cause signal happens to be flushed with the
  instruction; any future exception source derived from persistent state (debug triggers, etc.) will re-create this class
  of bug; (b) PCM of a flushed instruction is live garbage (ifu.sv:420 has no flush) — harmless only as long as nothing
  samples it without a valid trap, which is precisely the assumption that just failed. Neither needs fixing in this PR,
  but both deserve at least a comment, since the PR's correctness now silently depends on (a).

  4. Process gaps. No regression test is added; the fix is verified (presumably) only against the attached ELF. The fixing
  commit should state that InterruptsU lockstep now passes plus what regression was run. Minor: the new wfiM &
  ~WFICount[16] condition adds FEC expression rows on line 91 that only get hit when a timeout actually occurs — worth
  checking rv64gc --ccov results post-merge so this doesn't show up as a new uncovered row next to the existing exclusion
  block at privdec.sv:93.

rosethompson:
Hi @Teemooooooooo thank you for discovering this bug.   This exposes an interesting corner case(s).  I have three comments.
1. The first issuess is the wfi time out counter does not stop incrementing on the cycle it flags the time out as you described.  This is partial red herring, but should be corrected. Even if the counter stopped it at the correct time would continue to output the time out flag for an extra cycle. (i.e. it was held at 0x1000 for 2 cycles)
2. The real fix is two changes. First, the counter must stop incrementing when the time out flags and continue to hold its value.  This is vital.  If some other hazard stalls the CPU, the wfi time out must be held until that hazard is resolved and the instruction can trap.  Second, when the instruction traps, the counter must reset.
3. Claude thinks we need to mask ExceptionM with InstrValidM.  I'm not sure this is necessary, but it is probably ok. However I am against it's implementation. By construction the exceptions should not be possible to generate unless the instruction is valid.  wfi has this bug.  We could argue masking ExceptionM with InstrValidM may hide bugs when the driving fault should not be generated.  These faults fan out to other logic in the CPU.  I would rather the CPU fail with a trap than with a more subtle bug.

I believe the implementation fix is something like the following.

```
  if (P.U_SUPPORTED) begin : wfi
    logic [P.WFI_TIMEOUT_BIT:0] WFICount, WFICountPlus1;
    logic                       WFICountEn, WFICountRst;
    assign WFICountRst = reset | TrapM;
    assign WFICountEn = ~WFITimeoutM;
    assign WFICountPlus1 = wfiM ? WFICount + 1 : '0; // restart counting on WFI
    flopenr #(P.WFI_TIMEOUT_BIT+1) wficountreg(clk, WFICountRst, WFICountEn, WFICountPlus1, WFICount);  // count while in WFI
  // coverage off -item e 1 -fecexprrow 1
  // WFI Timeout trap will not occur when STATUS_TW is low while in supervisor mode, so the system gets stuck waiting for an interrupt and triggers a watchdog timeout.
    assign WFITimeoutM = ((STATUS_TW & PrivilegeModeW != P.M_MODE) | (P.S_SUPPORTED & PrivilegeModeW == P.U_MODE)) & WFICount[P.WFI_TIMEOUT_BIT];
  // coverage on
  end else assign WFITimeoutM = 1'b0;
```

Teemooooooooo:
Sounds good! I will push a new commit implementing this fix.

davidharrishmc (approved):

PR #1784, opened 2026-06-11
Fix cause mask for imperasfpm to accept causes above 15

davidharrishmc:


config/rv64gc/imperasfpm.ic
```diff
@@ -48,7 +48,7 @@
 --override iss/cpu0/Sstc=T
 
 # mcause and scause only have 4 lsbs of code and 1 msb of interrupt flag
```
jordancarlin: Comment needs to be updated.

config/rv64gc/imperasfpm.ic
```diff
@@ -48,7 +48,7 @@
...
---override iss/cpu0/ecode_mask=0x800000000000000F # for RV64
+--override iss/cpu0/ecode_mask=0x800000000000001F # for RV64
```
jordancarlin, line 51: Should this be done for rv32 too?

PR #1786, opened 2026-06-13
Fix code coverage tooling

davidharrishmc:
Detect livelock of trap handler trapping back to itself forever
Report tests that were silently failing when Imperas mismatched

regression-wally --ccov now finishes and generates report rather than hanging

### DIFF
--- a/testbench/common/watchdog.sv
+++ b/testbench/common/watchdog.sv
@@ -24,7 +24,7 @@
 // and limitations under the License.
 ////////////////////////////////////////////////////////////////////////////////////////////////
 
-module watchdog #(parameter XLEN, WatchDogTimerThreshold)
+module watchdog #(parameter XLEN, WatchDogTimerThreshold, SameTrapPCThreshold = 100)
   (input clk,
    input reset,
    string TEST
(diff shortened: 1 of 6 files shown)


testbench/common/watchdog.sv
```diff
@@ -44,11 +44,42 @@ module watchdog #(parameter XLEN, WatchDogTimerThreshold)
...
+  end
+
   always_comb begin
     WatchDogTimeOut = WatchDogTimerCount >= WatchDogTimerThreshold;
-    if(WatchDogTimeOut) begin
-      if (TEST == "buildroot") $display("Watch Dog Time Out triggered.  This is a normal termination for a full buildroot boot.  Check sim/<simulator>/logs/buildroot_uart.log to check if the boot printed the login prompt.");
-      else $display("FAILURE: Watch Dog Time Out triggered. PCW stuck at %x for more than %d cycles", PCW, WatchDogTimerCount);
+    SameTrapPCTimeOut = SameTrapPCCount >= SameTrapPCThreshold;
+    if(WatchDogTimeOut | SameTrapPCTimeOut) begin
+      if (TEST == "buildroot" & WatchDogTimeOut) $display("Watch Dog Time Out triggered.  This is a normal termination for a full buildroot boot.  Check sim/<simulator>/logs/buildroot_uart.log to check if the boot printed the login prompt.");
```
jordancarlin, line 80: This masks a livelock in buildroot. It might be worth distinguishing between the two timeouts.

jordancarlin (approved):
One minor comment. Otherwise, LGTM.

testbench/common/watchdog.sv
davidharrishmc: I think this is right.  Buildroot has a watchdog timeout when it gets to the login prompt, so that shouldn't fail.  We still want to report if Buildroot got in a trap loop.

jordancarlin: I missed the & WatchDogTimeOut. Agreed this is correct as is.

PR #1790, opened 2026-06-15
Add OpenSUSE 15.6 Workaround

fourth-bit:
https://bugzilla.suse.com/show_bug.cgi?id=1266957 was closed as WONTFIX. This provides a workaround that downgrades the necessary packages, so we can successfully install every necessary package.


jordancarlin (approved):
Seems hacky, but if that is the only way to make it work, then LGTM assuming CI passes.

PR #1791, opened 2026-06-16
Fix install issue with buildroot on Ubuntu 26.04

fourth-bit:
The buildroot installation was failing due to a problem with the version of `install` provided in Ubuntu 26.04. The suggested solution is to run `update-alternatives --install /usr/bin/install install /usr/bin/gnuinstall 100`, however I think that is not a suitable solution for our install scripts, as it replaces `install` with `gnuinstall` even after our install script finishes running. Instead, the solution here symlinks `gnuinstall` into a temporary directory that is placed into `PATH`  to make the buildroot installation works without having other side effects.

### DIFF
--- a/bin/installation/buildroot-install.sh
+++ b/bin/installation/buildroot-install.sh
@@ -44,7 +44,18 @@ STATUS="buildroot"
 export LD_LIBRARY_PATH=$RISCV/lib:$RISCV/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}:$RISCV/riscv64-unknown-elf/lib:$RISCV/lib/x86_64-linux-gnu
 cd "$WALLY"/linux
 if [ ! -e "$RISCV"/buildroot ]; then
-    FORCE_UNSAFE_CONFIGURE=1 make 2>&1 | logger; [ "${PIPESTATUS[0]}" == 0 ] # FORCE_UNSAFE_CONFIGURE is needed to allow buildroot to compile when run as root
+    if [ "$FAMILY" == ubuntu ] && ((UBUNTU_VERSION == 26)); then
+        # Buildroot currently does not work with the default version of install: https://github.com/uutils/coreutils/issues/12166
+        # The update-alternatives fix will not work here because it makes a permanent change
+        TMP_BIN=/tmp/riscv-bin
+        mkdir -p TMP_BIN
+        ln -sf "$(which gnuinstall)" $TMP_BIN/install
+        # FORCE_UNSAFE_CONFIGURE is needed to allow buildroot to compile when run as root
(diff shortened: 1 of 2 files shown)


jordancarlin:
This seems strange. CI for Ubuntu 26.04 passed on the PR that added it (https://github.com/openhwgroup/cvw/actions/runs/25936368403/job/76242761375) and I was able to successfully install on Ubuntu 26.04 locally. Did something change?

jordancarlin:
Looks like it was updated in the last few weeks (after CI had already run for the Ubuntu 26.04 PR): https://lists.buildroot.org/pipermail/buildroot/2026-June/804075.html. Based on the linked issue, I think we probably should go ahead and just use the command to replace it with the gnu version. Looks like this is a very significant bug in the uutils version.

bin/installation/buildroot-install.sh
```diff
@@ -44,7 +44,18 @@ STATUS="buildroot"
 export LD_LIBRARY_PATH=$RISCV/lib:$RISCV/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}:$RISCV/riscv64-unknown-elf/lib:$RISCV/lib/x86_64-linux-gnu
 cd "$WALLY"/linux
 if [ ! -e "$RISCV"/buildroot ]; then
-    FORCE_UNSAFE_CONFIGURE=1 make 2>&1 | logger; [ "${PIPESTATUS[0]}" == 0 ] # FORCE_UNSAFE_CONFIGURE is needed to allow buildroot to compile when run as root
+    if [ "$FAMILY" == ubuntu ] && ((UBUNTU_VERSION == 26)); then
+        # Buildroot currently does not work with the default version of install: https://github.com/uutils/coreutils/issues/12166
+        # The update-alternatives fix will not work here because it makes a permanent change
+        TMP_BIN=/tmp/riscv-bin
+        mkdir -p TMP_BIN
+        ln -sf "$(which gnuinstall)" $TMP_BIN/install
```
jordancarlin: Do we need to install `gnuinstall` first?

fourth-bit:
I feel like its not up to our install scripts to administrate other people's systems for them though. The other problem I have with update-alternatives is that it requires root access to run, which the buildroot installation does not require. Perhaps we should use the same check buildroot uses, issue a warning if its the wrong version of install, and then symlink gnuinstall into PATH?

bin/installation/buildroot-install.sh
fourth-bit: Its part of `gnu-coreutils` which is listed on the package manifest for all ubuntu 26.04 variants that I see on their website, and is in the ubuntu:26.04 container from podman

jordancarlin:
How about taking the same approach we use for gcc-10 on Ubuntu 20.04 in `activate-tools.sh`? Add another clause to that file for Ubuntu 26.04.

jordancarlin (approved):

PR #1795, opened 2026-06-24
Add errata entry for link correction on page 20 of RISC-V System-on-Chip Design

prithvidiamond1:
This change only affects the digital version of the book.

### DIFF
--- a/errata.md
+++ b/errata.md
@@ -6,6 +6,7 @@ Please contribute by making a pull request to modify this document on GitHub.  S
 
 | Page | Location | Error | Correction  | Contributor | Date |
 | ---- | -------- | ----- | ----------- | ----------- | ----|
+| 20 | 1.6 3rd paragraph | Link for Fig 1.21 takes user to section F.2.6 | fix the link | Prithviraj Prasad, USA | 6/23/26 |
 | 31 | Ch. 2 opening paragraph | "RISC processor sign" | Replace with "RISC processor design". | Sotaro Fujimoto, Japan | 6/17/26 |
 | 133 | 3.7 | The ACT suite has been completely redesigned and no longer uses RISCOF.  |See the [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test) repository for updated information.|David Harris, Claremont, CA | 1/4/26 |
 | 226 | 5.9.2 | cvw-arch-verif has been deprecated and no longer is in use | Replace with riscv-arch-test | David Harris, Claremont, CA | 1/4/26 |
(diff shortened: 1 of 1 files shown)


prithvidiamond1:
Ah shoot, I signed the ECA after I submitted this. Just rerun the workflow to verify this.

errata.md
```diff
@@ -6,6 +6,7 @@ Please contribute by making a pull request to modify this document on GitHub.  S
 
 | Page | Location | Error | Correction  | Contributor | Date |
 | ---- | -------- | ----- | ----------- | ----------- | ----|
+| 20 | 1.6 3rd paragraph | Link for Fig 1.21 takes user to section F.2.6 | fix the link | Prithviraj Prasad, USA | 6/23/26 |
```
jordancarlin: Can you please double check this? I just tested this with the WEB version of the book that I have, and the link for Fig 1.21 takes me to page 22, which is where that figure is.

prithvidiamond1: Oh, I wasn't aware that there was a WEB version (how come buyers of the book don't get that? I only have PDF and ePUB). I was referring to the PDF version. I should have been more specific. I tried it on both Chrome's PDF viewer and that of Firefox, and got the same result (and I believe they use different PDF viewers, so it's unlikely to be an implementation error of the PDF viewer).

jordancarlin: When I said WEB version, I was referring to the PDF version. Not sure why I am seeing a different result, but I am fine merging this if you are seeing this issue on two different PDF viewers. Can you update the "Correction" to mention the page that it should actually link to, and then this is good to go.

prithvidiamond1: I just did so. Also, if you want, I can send you my copy of the PDF (not here for obvious reasons, probably by email or something else that is suitable) so you can check if the problem shows up on your side. If it does, then it might be a problem on the side of your publisher or whoever is distributing the digital version of the book.

jordancarlin (approved):

errata.md
jordancarlin: No need. Thanks though.

prithvidiamond1: Thanks for merging this PR. I do have one last question, though. I am still reading through the book (it was recommended to me by a very good friend of mine), and it is possible I may come across more issues like this that may not be visible on your copy of the book. Would you prefer I report them by making more PRs or not?

jordancarlin: Please feel free to continue reporting any issues in the book that you come across.

prithvidiamond1: Will do then, thanks!

PR #1803, opened 2026-07-17
PWM Peripheral

naichewa:
Adds a PWM peripheral with 4 16 bit comparators and 2 tests. WALLY-pwm-01.S contains "dumb" functions that wait the estimated duration of X pwm cycles, allowing for tests by inspection. WALLY-pwms-01.S counts the number of high/low cycles and compares to reference


tests/wally-riscv-arch-test/riscv-test-suite/rv32i_m/privilege/src/WALLY-pwm-01.S
```diff
@@ -0,0 +1,235 @@
+///////////////////////////////////////////
+//
+// WALLY-pwm
+//
+// Author: Naiche Whyte-Aguayo naichewa@gmail.com>
```
davidharrishmc, line 5: Tests should have some introductory text explaining what they are testing.  The reference_output of pwm-01.S suggests it isn't testing much.  Is it worth keeping this one?  

Most of the test is commented out.  Is it worth keeping the commented out parts?  If so, explain why they are there.

All the features of the module should be fully tested in a self-checking fashion.

Same comments on pwms.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
+///////////////////////////////////////////
+// pwm_apb.sv
+//
+// Written: naichewa@gmail.com 4/29/2026
+// Modified:
+//
+// Purpose: PWM module
+//  https://docs.google.com/document/d/1erHBVchBtwmgZ0bCNjb88spYfN7CpRbhmSNFH6cO8CY/edit?tab=t.0
+//
+// Documentation: RISC-V System on Chip Design
```
davidharrishmc: Not documented in the book.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
+///////////////////////////////////////////
+// pwm_apb.sv
+//
+// Written: naichewa@gmail.com 4/29/2026
+// Modified:
+//
+// Purpose: PWM module
+//  https://docs.google.com/document/d/1erHBVchBtwmgZ0bCNjb88spYfN7CpRbhmSNFH6cO8CY/edit?tab=t.0
```
davidharrishmc: This URL doesn't describe the PWM.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+//
+// Documentation: RISC-V System on Chip Design
+// Based on PWM design from SiFive 5U540-C000 manual version 1.0
```
davidharrishmc: Is this the FU540 chip, not 5U540?  Include a link i possible, such as https://pdos.csail.mit.edu/6.828/2025/readings/FU540-C000-v1.0.pdf
or one to the SiFive site if available.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+  assign PWMCompareCenter = PWMConfig[12:9];
+  assign PWMCompareGang = PWMConfig[16:13];
+
+  assign Carryout = &PWMScaled;
+  assign PWMCountEn = PWMEnAlways | PWMEnOneShot;
+  assign PWMCountIncrement = PWMCount + 1;
+  assign PWMCountPrescaled = ((Memwrite & (Entry == PWM_COUNT)) ? Din[30:0] : PWMCountIncrement) >> PWMScale;
+
+
+  // Combinatorial signal logic
```
davidharrishmc: Combinational

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+      PWMCompare0 <= 16'hFFFF;
+      PWMCompare1 <= 16'hFFFF;
+      PWMCompare2 <= 16'hFFFF;
+      PWMCompare3 <= 16'hFFFF;
+    end else begin // writes
+      /* verilator lint_off CASEINCOMPLETE */
+      if (Memwrite)
+        case(Entry) // flop to sample inputs
+          PWM_CFG:   PWMConfig <= {Din[31:24], Din[19:16], Din[13:12], Din[10:8], Din[3:0]};
+          //PWM_COUNT: PWMCount <= Din[30:0];
```
davidharrishmc: Delete unused signals

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+  // PWMCount and PWMScaled registers
+
+  always_ff @(posedge PCLK)
+    if (PWMCountReset | ~PRESETn) begin
+      PWMCount <= 31'b0;
+      PWMScaled <= 16'b0;
+    end else if (Memwrite & (Entry == PWM_COUNT)) begin
+      PWMCount <= Din[30:0];
+      PWMScaled <= PWMCountPrescaled[15:0];
+    end else if (Memwrite & (Entry == PWM_S))PWMScaled <= Din[15:0];
```
davidharrishmc: Space after paren

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+    else if (PWMCountEn) begin
+      PWMCount <= PWMCountIncrement;
+      PWMScaled <= PWMCountPrescaled[15:0];
+    end
+
+  flop #(1) pwmholdreg(PCLK,
```
davidharrishmc: weird line break. Several places

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+                       PWMHoldIn, PWMHoldOut);
+
+  //pwm comparators
+  assign PWMDeglitchMux[0] = PWMScaled[15] & PWMCompareCenter[0];
```
davidharrishmc: This needs a better name. It is not a mux, and is also an input to an xor.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+  assign PWMCompareXNOR[15:0] = PWMDeglitchMux[0] ? ~PWMScaled : PWMScaled;
+  assign PWMCompareBoolean[0] = PWMCompareXNOR[15:0] >= PWMCompare0;
```
davidharrishmc: Rename.  Boolean isn't a helpful name for a binary signal.

src/uncore/pwm_apb.sv
```diff
@@ -0,0 +1,233 @@
...
+      endcase
+    end
+
+
+  // PWMCount and PWMScaled registers
+
+  always_ff @(posedge PCLK)
+    if (PWMCountReset | ~PRESETn) begin
+      PWMCount <= 31'b0;
+      PWMScaled <= 16'b0;
```
davidharrishmc: I'm not clear whether pwmscaled needs to be a separate 16-bit register, or if it can just be the result of combinational logic shifting pwmcount by up to 15.  Although it is drawn as a register in Fig 6, it looks like it could be combinational because it doesn't hold any new state.

davidharrishmc (changes requested):

src/uncore/pwm_apb.sv
naichewa: Spec has it as an acessible read/write-able register, although I agree other than following that it could be combinational. I also don't see why you would want to write to it and not PWM Count

davidharrishmc:
AI Review:

A faithful structural port: the register map, `pwmcfg` field layout, counter/scaler, and all four waveform
modes (left/right-align, center, ganging, one-shot) match the spec. But the **interrupt-pending
(`pwmcmpXip`) path has two functional bugs** and there is **one counter-reset deviation**, so it is not yet
spec-compliant for interrupt or timer use. Fix those before merge.

## Conforms to FU540 §14

- Memory map: `0x00 pwmcfg, 0x08 pwmcount, 0x10 pwms, 0x20/24/28/2C pwmcmp0–3` (Table 68); base
  `0x10020000` = FU540 PWM0 (Table 67).
- `pwmcfg` bit layout (Table 70): scale[3:0], sticky 8, zerocmp 9, deglitch 10, enalways 12, enoneshot 13,
  cmp0–3center 16–19, cmp0–3gang 24–27, cmp0–3ip 28–31. The write path (line 138) compacts the sparse
  layout into a 21-bit register and the read path (line 147) re-expands it with the reserved gaps; field
  positions are correct.
- `pwmcount` = cmpwidth+15 bits, bit 31 reads 0 for cmpwidth=16 (§14.4). `pwms = pwmcount >> pwmscale`
  truncated to cmpwidth (§14.5/14.6).
- Comparator `pwms >= pwmcmpX` (§14.7); center mode inverts pwms when its MSB is set (§14.10, Table 76);
  ganging `gpio[i] = ip[i] & ~(gang[i] & ip[i+1])` with 3→0 wraparound (§14.11); one-shot self-clears
  `pwmenoneshot` on the reset condition (§14.12).
- Reset: `pwmcfg=0` (enables off, per Table 70 rst 0x0). Compare regs reset to all-1s (spec says X — a
  harmless, arguably safer choice).
- Integration: `adrdec` region, `PWMGPIO`→GPIO iof0[3:0], PLIC request ID 7, `PWM_WIDTH=16`, config
  plumbing across all targets, and a disabled-config stub — all clean.

## Bugs and deviations (most severe first)

### 1. `pwmcfg` reads back the *written* ip bits, not the live pending flops — functional bug
`pwm_apb.sv:147`
```
PWM_CFG: Dout <= {PWMConfig[20:13], 4'b0, PWMConfig[12:9], 2'b0, PWMConfig[8:7], 1'b0, PWMConfig[6:4], 4'b0, PWMConfig[3:0]};
```
`PWMConfig[20:17]` holds whatever software last wrote to `pwmcfg[31:28]` (from the write pack at line 138,
`Din[31:24]`). The true interrupt-pending status lives in the independent `PWMCompareIP[3:0]` flops (line
173–197). So a read of `pwmcfg[31:28]` never reflects `pwmcmpXip`; software polling `pwmcfg` for pending
interrupts reads stale/garbage bits.
**Fix:** drive `pwmcfg[31:28]` from `PWMCompareIP` on read, not `PWMConfig[20:17]`.

### 2. `pwmcmpXip` cannot be cleared by writing `pwmcfg` — functional bug
Spec §14.13: *"The interrupt pending bits pwmcmp ip can be cleared down using writes to the pwmcfg
register."* There is no write path to the `PWMCompareIP` flops — they are driven only by the
comparator/deglitch mux (`pwm_apb.sv:171/179/187/195`). Consequently the interrupt (`PWMIntr =
|PWMCompareIP`, line 200) cannot be acknowledged. With **`pwmsticky=1`** it is worse:
`PWMHoldIn = (~PWMOneShotEnReset & PWMDeglitch) | PWMSticky` (line 112) forces `PWMHoldOut` high, so the
mux is `comp | ip` and a set ip **latches permanently** — the exact case §14.13 says sticky is for
(ensuring interrupts "are not forgotten"), but with no clear path it can never be serviced.
**Fix:** add a write-1-to-clear (or write-value) path from a `pwmcfg` write into `PWMCompareIP`, gated so
`pwmsticky` blocks clearing only per the spec ("disallow clearing if already set" is about the hardware
re-clear, while software clears via the register write).

### 3. Counter resets on `Carryout` even when `pwmzerocmp=0` — spec deviation
`pwm_apb.sv:105–108`
```
assign Carryout          = &PWMScaled;                               // pwms all ones
assign PWMOneShotEnReset = (PWMComparator[0] & PWMZeroCompare) | Carryout;
assign PWMCountReset     = PWMOneShotEnReset | (~PRESETn);
```
`Carryout` unconditionally resets `pwmcount`. Spec §14.4/§14.13: with `pwmzerocmp=0` the counter *"is
simply allowed to wrap around to zero"* and the block *"can also be used as a regular timer with **no
counter reset** (pwmzerocmp=0)."* Here `pwmcount` instead resets every pwms cycle, so it never exceeds
`2^(cmpwidth+scale)` and the documented free-running cmpwidth+15-bit timer is truncated. Harmless for pure
waveform output (pwms is identical either way), wrong for the timer/`pwmcount`-read use case. In §14.8
`Carryout` (pwms all-ones) is defined as the **deglitch trigger** when `pwmzerocmp=0`, not a count reset.
**Fix:** feed `Carryout` to the deglitch logic only; gate it out of `PWMCountReset` when `pwmzerocmp=0`
(let the counter free-run/naturally wrap).

### 4. Single ORed PLIC interrupt vs. four per-comparator sources — simplification
FU540 routes `pwmcmp0ip…pwmcmp3ip` as four independent PLIC sources; the PR ORs them into one `PWMIntr`
(`pwm_apb.sv:200`, PLIC ID 7). Acceptable for Wally, but combined with bugs 1–2 the handler can neither
identify nor clear the specific firing comparator. Worth documenting as an intentional deviation.

## Test caveat

The PR adds `WALLY-pwm-01.S` and reference outputs for rv32/rv64, but the references are generated from this
same RTL, so a passing test is self-consistent, not spec-proof — and would bake in the ip-readback/clear
behavior above. Not built/run for this review. The test should be extended to (a) read `pwmcfg` and check
`[31:28]` reflects live `pwmcmpXip`, and (b) set a comparator, take/observe the interrupt, clear it via a
`pwmcfg` write, and confirm it clears (and that `pwmsticky` behaves per §14.13).

davidharrishmc:
Merged, for Claude to touch up.

PR #1826, opened 2026-09-02
Add Table 10.2 ClearDirty erratum

thypled:
Table 10.2 on page 455 lists the ClearDirty flush criterion as `(Flush & ~LineDirty)`. It should be `(Flush & LineDirty)`.

This matches Section 10.4.5.4, which states that ClearDirty is asserted on the transition from Flush to FlushWriteback, and the condition in `cachefsm.sv`.

Reported by 王晗宇, China.


davidharrishmc (approved):

davidharrishmc:
Thank you!

PR #1829, opened 2026-09-04
Convert peripheral tests to self-checking ELFs built without RISCOF

davidharrishmc:
## Summary

Converts the wally-riscv-arch-test peripheral tests (uart, gpio, clint, plic, plic-s, periph, periph-s, spi, uart-timeout) into self-checking ELFs that are built without RISCOF, as a first step toward retiring the RISCOF flow.

- Each test in `tests/periph/src` embeds its expected signature and checks every entry as soon as it is written. A mismatch records the entry index, address, expected and actual values in `selfcheck_record` and halts; the testbench's new `CheckSelfCheck` task prints it in the familiar `Error on test ... result N` format. No reference files or signature diffs.
- `tests/periph` is self-contained: `WALLY-env.h` replaces the riscv-arch-test framework headers, `WALLY-TEST-LIB.h` is one trimmed library serving both widths (unused macros, test types and virtual-memory plumbing removed), and `link.ld` is local. `make periph` (also run by the top-level `make`) builds every source for RV32 and RV64 into `rv32/` and `rv64/`.
- One source per test works for both widths through `LREG`/`SREG`/`WORD`/`REGBYTES` and `SEXT32()` for expected values that are sign extended on RV64. `WALLY-periph-01.S` with `-DM_MODE_ONLY` provides the rv32imc variant (`WALLY-periph-m-01`).
- Suites: `wally32periph` (rv32gc, 7 tests), `wally32periph_imc` (rv32imc), `wally64periph` (7 tests). The rv32 peripheral tests are removed from `wally32priv`. The RISCOF PWM test stays in `wally32priv`/`wally64priv`; it is converted in a follow-up PR.

Notes:
- The old reference files were missing the final terminate-ecall entry (the old flow never compared past the end of the reference); the embedded tables include it.
- `WALLY-uart-timeout-01` is built but not in a suite because it fails on the current design (the character-timeout interrupt never appears).

## Test plan

- [x] `wally32periph`, `wally32periph_imc`, `wally64periph` pass on Verilator; `wally32periph`/`wally64periph` pass on Questa
- [x] Deliberately wrong expected value and wrong entry count are reported with index, address, expected and actual
- [x] One injected RTL bug per peripheral (uart MSR, gpio rise_ip, clint MSIP, plic threshold, spi sckdiv) is caught by the suites
- [x] Unconverted RISCOF tests assemble to identical code (their library is untouched)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1830, opened 2026-09-04
Fix PWM pending bits, counter reset, and pwms write; add thorough self-checking PWM test

davidharrishmc:
## Summary

Stacked on #1829 (the first commit here is that PR; review only the second commit).

Fixes in `src/uncore/pwm_apb.sv`, checked against the FU540-C000 manual chapter 14, covering the issues found by the new test and by the review of #1803:

- `pwmcfg[31:28]` reads the live `pwmcmpXip` flops instead of the last written configuration bits.
- A `pwmcfg` write sets the `pwmcmpXip` flops to the written bits, so software can clear a pending interrupt (or set one). `pwmsticky` only stops the hardware from clearing, as the manual describes.
- The counter resets only when `pwmzerocmp` is set and counting; otherwise it free-runs and wraps naturally so it can serve as a timer. The pwms wrap still ends a one-shot cycle and releases the deglitch hold.
- The pwms write is widened before shifting by `pwmscale`; previously the shift happened in a 16-bit expression and truncated the value for nonzero scales.
- Each comparator drives its own PLIC interrupt, as in the FU540, replacing the single ORed source. The ORed source was raised whenever any `pwmcmpXip` was set, so it fired at the `pwms` wrap even when the comparator software cared about was low: with comparator 0 centered and the others at all ones, the wrap raised the PWM interrupt although comparator 0 was low. Software also no longer has to read `pwmcfg[31:28]` to find out which comparator fired.

The four comparators need four PLIC source IDs, and no four contiguous IDs were free below 10, so `PLIC_NUM_SRC` grows from 10 to 14 and the PWM block moves from source 7 to 11-14. GPIO, SPI, SDC and UART keep their IDs, which matters because `linux/devicetree/wally-*.dts` pins GPIO at 3 and UART at `0x0a`; PWM appears in no device tree and `PLIC_PWM_ID` had no user outside `plic_apb.sv`, so nothing else moves. Software gets the comparator index as `claimID - PLIC_PWM_ID`.

In the test library, `SETUP_PLIC` raises the priority of all four PWM sources rather than just source 7, and `claim_m_plic_interrupts` learns the four new priority registers, widens the enable mask it installs from `0xFFF` to `0x7FFF` so sources 12 through 14 can be claimed, and drains every pending source instead of one. The drain is what the split requires: several comparators can be pending at once, and a single claim left `mip` still showing the external interrupt. It is bounded at 16 so a still-asserted source cannot hang a test.

The existing checks could not see the split at all. Every interrupt observation was either `mip` — a single external interrupt bit set by any pending source — or `pwmcfg[31:28]`, which reads the pending flops inside the PWM, upstream of the PLIC; `claim_m_plic_interrupts` wrote nothing to the signature. Driving all four sources from the ORed value still passed the whole suite. A new test type, `claim_m_plic_id_test`, claims one interrupt and records the source ID, and three new sections use it: each comparator raises its own source and no other; comparators pending together are claimed one per source in increasing order, with `mip` staying set in between; and a wrap with comparators 0, 2 and 3 centered and comparator 1 at all ones reports comparator 1 alone — Naiche's example, the case the ORed source got wrong. With the ORed behaviour restored the suite now fails.

`WALLY-pwm-01` moves to `tests/periph/src` as one self-checking source for RV32 and RV64, and the RISCOF copies and references are removed. Its wait routine now synchronizes on the first pulse and observes N - 1/2 further periods, so pulse counts no longer depend on software loop timing (the original exact-N-period window read 3 pulses instead of 4 on main). New sections cover register widths and readback, the `pwmcfg` pending bits, per-comparator timer-mode interrupts, deglitch in wrap mode, center mode on all comparators and with deglitch, all four gang pairs checked by output level, one-shot self-clearing and replay, a 100% duty output, the free-running counter past the pwms wrap, and software set/clear of pending bits. Signature entries: 30 to 105. All checks are register values, output levels, or pulse counts; none depend on edge positions.

## Test plan

- [x] `wally32periph` and `wally64periph` (including the PWM test) pass on Verilator and Questa; `wally32periph_imc` passes on Verilator
- [x] Restoring the ORed interrupt makes the new per-source checks fail, so the split is covered
- [x] `WALLY-plic-01` and `WALLY-plic-s-01` pass unchanged with `PLIC_NUM_SRC` at 14
- [x] `lint-wally` clean
- [x] Before the fixes, six of the new checks failed on the merged RTL (pwms write truncation, pending-bit readback); all pass after
- [x] An injected comparator bug (`>=` to `>`) is still caught by the timing-independent wait routine


davidharrishmc:
After 1829 is merged, send this one to @naichewa for review.

tests/periph/WALLY-TEST-LIB.h
```diff
@@ -734,6 +750,112 @@ trap_handler_end_\MODE\(): // place to jump to so we can skip the trap handler a
...
+    sll t2, t3, t4 // N periods
+    srli t6, t3, 1 // half a period
+    sub a5, t2, t6 // window after the first edge: N - 1/2 periods
+    add t2, t2, t3 // N + 1 periods: limit for the first edge
+    rdcycle a4
+    add a4, a4, t2 // time limit for the first edge
+    li t4, 0x1006002C // GPIO high_ip address
+    li t5, 0 // number of high intervals seen
+    li t6, 0xFFFFFFFF
+    sw t6, 0(t4) // clear high_ip
```
naichewa, line 978: Doesn't end up waiting integer*FFFF clock cycles for center tests. Example, assume want to wait FFFF*1.
in pwm_cycle_wait_common, FFFF > 1 = 7fff, window after first edge  FFFF-7fff =8000. With an early first edge, can miss the transition from high to low

src/uncore/pwm_apb.sv
```diff
@@ -169,34 +175,31 @@ module pwm_apb import cvw::*; #(parameter cvw_t P) (
...
-  flop #(1) pwmcompareipreg3(PCLK,
-                            PWMDeglitchMux[3], PWMCompareIP[3]);
   assign PWMGPIO[3] = PWMCompareIP[3] & ~(PWMCompareGang[3] & PWMCompareIP[0]);
 
+  // pwmcmpXip flops: a pwmcfg write sets them to the written value, otherwise the comparator and deglitch/sticky logic drives them
+  assign PWMCompareIPNext = PWMConfigWrite ? Din[31:28] : PWMDeglitchMux;
+  flopr #(4) pwmcompareipreg(PCLK, ~PRESETn, PWMCompareIPNext, PWMCompareIP);
+
+  // All four pending bits are ORed into one PLIC source (FU540 has four); software identifies the comparator through pwmcfg[31:28]
   assign PWMIntr = |(PWMCompareIP);
```
naichewa, line 203: ORed interrupts can cause interrupt to fire when none is wanted. Example, comparator 0 in center mode is 0 at pwms count wrap of ffff but other comparators set to max will cause an interrupt

src/uncore/pwm_apb.sv
```diff
@@ -102,20 +104,24 @@ module pwm_apb import cvw::*; #(parameter cvw_t P) (
...
-  assign Carryout = &PWMScaled;
+  assign Carryout = &PWMScaled; // pwms is all ones: the counter is about to wrap
   assign PWMCountEn = PWMEnAlways | PWMEnOneShot;
-  assign PWMOneShotEnReset = (PWMComparator[0] & PWMZeroCompare) | Carryout;
-  assign PWMCountReset = PWMOneShotEnReset | (~PRESETn);
+  // A PWM cycle ends when pwms reaches pwmcmp0 with pwmzerocmp set, or when pwms wraps.  The end of a cycle
+  // finishes a one-shot and releases the deglitch hold.  Only pwmzerocmp resets the counter; otherwise it
+  // free-runs and wraps naturally so it can serve as a timer (FU540 manual 14.4, 14.13).
+  assign PWMCycleEnd = PWMCountEn & ((PWMComparator[0] & PWMZeroCompare) | Carryout);
+  assign PWMCountReset = PWMCountEn & PWMComparator[0] & PWMZeroCompare;
```
naichewa, line 113: CountReset can be low at Carryout, allowing PWMCount to continue. Seems to contradict FU540 14.1 figure 6. Its ambiguous whether Carryout comes from PWMS or PWMCount, but section 14.8 states "if pwmdeglitch is set, but pwmzerocmp is cleaer, the deglitch circuit... is now triggered when pwms contains all 1s and will cause a carry out of the high bit of the pwms incrementer just before the counter wraps to zero" which supports pwms to me

naichewa (commented):
Possible discrepancies are:
pwm_cycle_wait_center doesn't wait a correct amount, window after first edge = (N = FFFF) - 1/2 = 8000. If first edge is fairly early, ex 00FF, window = 80FF and will cut off before the center wave form drops. Passes the number of highs, but waveform doesnt drop

Count is allowed to free run after pwms wraps. Seems to be for timer purposes?, but manual figure 6 in 14.1 shows pwms carryout connected to a PWMCount reset.

Single ORed PLIC source means interrupt fires at pwms wrap at FFFF, so interrupt can fire in center mode when observed comparator is low.

tests/periph/WALLY-TEST-LIB.h
davidharrishmc: Fixed with https://github.com/openhwgroup/cvw/pull/1830/changes/b4d6f2acd66376ad86f98609a9b73629f7d4f81f

src/uncore/pwm_apb.sv
davidharrishmc: Let's leave a single interrupt source for PWM to PLIC.

src/uncore/pwm_apb.sv
davidharrishmc: Fixed test and RTL with commit 33dea46

davidharrishmc:
Updated with your comments.  Could you give it another review and approve if you agree?

naichewa (commented):
Checked the waveforms and all looks good now, agree with the pwms carryout change.

Point of disagreement: I think the PWMInterrupt should match the FU540 spec, with each comparator having its own interrupt source instead of current single ORed interrupt. As is, the ORed interrupt is guaranteed to fire at wrap, even in situations where the intended comparator is low. For example, imagine cmp0 in centered mode, which will be low at wrap. All other comparators set to FFFF will fire at wrap and lead to an erroneous PWM interrupt.

davidharrishmc:
@naichewa, that makes sense. Implemented four PWM Interrupts.

naichewa (approved):

PR #1842, opened 2026-09-06
Fix two sh keywords in the csh setup scripts (fi, elseif)

tropibyte:
`setup.csh` closes an `if` with `fi`, and `site-setup.csh` uses `elseif`. Neither is csh, and tcsh does not accept either.

### `setup.csh:40` — `fi` should be `endif`

This is the damaging one. The block has an `else`, so after running either branch tcsh searches for `endif`, runs off the end of the file, and reports:

```
if: then/endif not found
```

(tcsh `sh.func.c:975`.) The practical effect is that **`setup.csh` only completes when `$RISCV/site-setup.csh` is absent** — when it *is* present, the `source` succeeds and then setup fails at the very last line.

### `site-setup.csh:57` — `elseif` should be `else if`

Quieter, but not harmless. tcsh parses `elseif ( ... )` as a command with a parenthesised argument and answers

```
Badly placed ()'s.
```

with no line number, which is why it is easy to miss. The `gcc-10` fallback branch is never taken, so on Ubuntu 20.04 the older toolchain is not placed on `PATH`.

### The change

Two tokens:

```diff
-fi
+endif
```
```diff
-elseif ( -e $RISCV/gcc-10 ) then
+else if ( -e $RISCV/gcc-10 ) then
```

### How this was found

Parsing a 1,287-script corpus of real-world csh/tcsh from 184 repositories with [cshw](https://github.com/tropibyte/CSHW), a csh implementation for Windows. Both files were verified to parse cleanly after the change.

Happy to split this into two PRs if you would rather take them separately.


davidharrishmc:
Thank you for this improvement.
You'll need to sign the eclipse ECA to make a open-source contribution.

Create an account with https://www.eclipse.org/
Sign the agreement, https://www.eclipse.org/legal/ECA.php

davidharrishmc (approved):

davidharrishmc:
Do you still want to proceed with this PR?

PR #1850, opened 2026-09-10
Vector FP fma lane

georgiatai:
Added a vector FP fma lane, covers add/sub/mul/fma, widening and non-widening:
- used existing unpack unit to direct the operand widths
- used existing FMA unit to complete arithmetics
- one post-processing unit in each lane in stead of one to avoid back-pressure

TODO:
- some widths are still determined by FLEN, kept for now for easier testing and waiting on ELEN support, would need to add such support to the existing fpu modules too
- can maybe make modifications to post-processing unit later on to eliminate unnecessary wires if chose to keep post-processing unit in each lane


georgiatai:
Will rebase to the current vector branch base, remove the post-processing unit from the lane for now, then reopen

georgiatai:
Kept post-processing unit in the file for now. As of right now I think it's better to keep the post-processing unit as a proof of concept of widening instructions. This can later on develop into a lane with multiple arithmetic blocks.

Issue #1839, opened 2026-09-06
Review of LSU vs. Spec


davidharrishmc:
# Review of `src/lsu` Against the Ratified RISC-V Specifications

**Date:** 2026-09-06
**Scope:** All ten files in `src/lsu/` (`lsu.sv`, `align.sv`, `amoalu.sv`, `atomic.sv`, `lrsc.sv`, `dtim.sv`, `endianswap.sv`, `subwordread.sv`, `subwordwrite.sv`, `swbytemask.sv`), plus the code the LSU's behavior depends on: the load/store/AMO/LR/SC/CBO/fence decode in `src/ieu/controller.sv`, `src/hazard/hazard.sv`, `src/privileged/trap.sv` (exception priority), `csr.sv`/`csrm.sv`/`csrs.sv`/`csrsr.sv` (xtval, envcfg CBIE/CBCFE/CBZE, MBE/SBE/UBE, MPRV), the misaligned-fault logic in `src/mmu/mmu.sv` and PMA size decode in `src/mmu/adrdecs.sv`, the CMO paths in `src/cache/cache.sv`, `cachefsm.sv`, `cacheway.sv`, and the uncached bus paths in `src/ebu/ahbcacheinterface.sv`, `buscachefsm.sv`, `busfsm.sv`. Translation, permission, and PMP/PMA matching were reviewed separately in `mmu_spec_review.md` and are referenced, not repeated.
**Reference:** Ratified spec text in `~/repos/riscv-isa-manual/src/`: `unpriv/rv32.adoc` (loads/stores, misaligned rules, endianness, FENCE), `unpriv/rv64.adoc`, `unpriv/a-st-ext.adoc`, `unpriv/zalrsc.adoc`, `unpriv/zaamo.adoc`, `unpriv/cmo.adoc` (Zicbom/Zicboz/Zicbop), `unpriv/d-st-ext.adoc`; `priv/machine.adoc` (mstatus MPRV/MBE/SBE/UBE, synchronous exception priority, mtval, PMA idempotency and misaligned-atomicity-granule, PMP decomposition of misaligned accesses), `priv/supervisor.adoc` (stval, sstatus.UBE). Zabha, Zacas, and Zawrs are not implemented in this tree (no references in `src/`), so their spec files were not applied.
**Baseline:** branch `cmpnv`, commit `1f56be750`.

## Summary

[...]

PR #1857, opened 2026-09-12
Fix LSU and CMO issues from #1839 (items 1-5)

davidharrishmc:
Addresses items 1-5 of #1839.

  **1. `cbo.inval` under `xenvcfg.CBIE=01` performed a clean, not a flush.** Moved the
  `CBIE=01` term from `CMOpD[1]` to `CMOpD[2]` so the downgraded invalidate takes the
  write-back-and-invalidate path. Previously the dirty line was written back but left
  valid, so a later load could never observe a store by a non-coherent agent. Corrected
  `CBIE=01` term from `CMOpD[1]` to `CMOpD[2]` so the downgraded invalidate takes the
  write-back-and-invalidate path. Previously the dirty line was written back but left
  valid, so a later load could never observe a store by a non-coherent agent. Corrected
  the swapped `CMOpM` bit-label comments (`1: cbo.inval; 2: cbo.clean; 4: cbo.flush;
  8: cbo.zero`) across `controller.sv`, `ieu.sv`, `lsu.sv`, `cache.sv`, `cachefsm.sv`,
  and `wallypipelinedcore.sv`.

  **2. CMO and uncached-AMO datapaths were gated on `ZICBOZ_SUPPORTED`.** Each is now
  gated on the extension it actually needs: `cbo.zero` to uncached memory on Zicboz, the
  cache CMO port on Zicbom or Zicboz, uncached AMOs on Zaamo. Added `zicbom_*`/`zicboz_*`
  derivatives of rv32gc and rv64gc so the split configurations are covered.

  `adrdecs.sv` reports those widths unsupported, so the access takes an access fault.

  **4. Misaligned accesses to non-idempotent regions raised address-misaligned.**
  `MisalignedFaultAllowedM` now also requires `Idempotent`, and
  `MisalignedCausesAccessFaultM` raises an access fault instead. Per `machine.adoc`,
  misaligned accesses to non-idempotent regions should raise access-fault rather than
  address-misaligned exceptions, so software does not emulate them with multiple smaller
  accesses that could have side effects. This puts the PMA `Idempotent` output to use, so
  its coverage exclusion is dropped.

  **5. Uncached `lbu`/`lhu`/`lwu` drove an illegal `HSIZE`.** AHB defines `HSIZE`
  100/101/110 as 128/256/512-bit transfers, but funct3 uses those codes for the unsigned
  loads. New `LSUSizeM` clears the unsigned bit for integer accesses, leaving FP accesses
  on their true width.

  Also in `controller.sv`, two fixes to `CMOFunctD` found while reviewing item 1:
  `P.ZICBOM_SUPPORTED` did not gate the `cbo.clean`/`cbo.flush` term due to operator
  precedence, so on a config without Zicbom those instructions decoded as legal no-ops in
  M-mode instead of raising an illegal instruction; and the `cbo.inval` CBIE test now
  enumerates `01`/`11` rather than `!= 00`, so the reserved `10` encoding cannot decode as
  a silent no-op.

### DIFF
--- a/src/lsu/lsu.sv
+++ b/src/lsu/lsu.sv
@@ -41,7 +41,7 @@ module lsu import cvw::*;  #(parameter cvw_t P) (
   input  logic [6:0]              Funct7M,                              // Atomic memory operation function
   input  logic [1:0]              AtomicM,                              // Atomic memory operation
   input  logic                    FlushDCacheM,                         // Flush D cache to next level of memory
-  input  logic [3:0]              CMOpM,                                // 1: cbo.inval; 2: cbo.flush; 4: cbo.clean; 8: cbo.zero
+  input  logic [3:0]              CMOpM,                                // 1: cbo.inval; 2: cbo.clean; 4: cbo.flush; 8: cbo.zero
   input  logic                    LSUPrefetchM,                         // Prefetch; presently unused
   output logic                    CommittedM,                           // Delay interrupts while memory operation in flight
   output logic                    SquashSCW,                            // Store conditional failed disable write to GPR
@@ -106,6 +106,7 @@ module lsu import cvw::*;  #(parameter cvw_t P) (
   logic [1:0]            PreLSURWM;                              // IEU or HPTW Read/Write signal
   logic [1:0]            LSURWM;                                 // IEU or HPTW Read/Write signal gated by LR/SC
   logic [2:0]            LSUFunct3M;                             // IEU or HPTW memory operation size
+  logic [2:0]            LSUSizeM;                               // Memory operation size as an AHB HSIZE
   logic [6:0]            LSUFunct7M;                             // AMO function gated by HPTW
   logic [1:0]            LSUAtomicM;                             // AMO signal gated by HPTW
   logic [3:0]            LSUCMOpM;                               // CMOpM gated by HPTW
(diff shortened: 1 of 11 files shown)


src/lsu/lsu.sv
```diff
@@ -346,7 +351,7 @@ module lsu import cvw::*;  #(parameter cvw_t P) (
         .HRDATA, .HWDATA(LSUHWDATA), .HWSTRB(LSUHWSTRB),
         .HSIZE(LSUHSIZE), .HBURST(LSUHBURST), .HTRANS(LSUHTRANS), .HWRITE(LSUHWRITE), .HREADY(LSUHREADY),
         .BeatCount, .SelBusBeat, .CacheReadDataWordM(DCacheReadDataWordM[P.LLEN-1:0]), .WriteDataM(LSUWriteDataM),
-        .Funct3(LSUFunct3M), .HADDR(LSUHADDR), .CacheBusAdr(DCacheBusAdr), .CacheBusRW, .BusAtomic, .BusCMOZero, .CacheableOrFlushCacheM,
+        .Funct3(LSUSizeM), .HADDR(LSUHADDR), .CacheBusAdr(DCacheBusAdr), .CacheBusRW, .BusAtomic, .BusCMOZero, .CacheableOrFlushCacheM,
```
rosethompson: The ahbcacheinterface should rename the Funct3 input to something else.  LSUSize does not work because this duplicated for the IFU.  I suggest Size.

src/mmu/adrdecs.sv
```diff
@@ -37,17 +37,21 @@ module adrdecs import cvw::*;  #(parameter cvw_t P) (
 );
 
   localparam logic [3:0]       SUPPORTED_SIZE = (P.LLEN == 32 ? 4'b0111 : 4'b1111);
+  // Uncached regions are accessed with a single AHBW-wide beat, so they cannot support an access wider
```
rosethompson: I don't think this comment is necessary.

rosethompson: Actually the name UNCACHED_SIZE is confusing. We are selecting which access widths are supported for clint and sdc peripherals with this localparam. A better name would be SUPPORTED_BUS_SIZE.

rosethompson:
I have a few requested changes, but other than the name change I am happy to merge.

davidharrishmc:
I can't find your requested changes.  Did you submit your review?

rosethompson (changes requested):

src/lsu/lsu.sv
davidharrishmc: Done

src/mmu/adrdecs.sv
davidharrishmc: Done

Issue #1838, opened 2026-09-06
Review of IFU vs. Spec


davidharrishmc:
# Review of `src/ifu` Against the Ratified RISC-V Specifications

**Date:** 2026-09-06
**Scope:** All files in `src/ifu/`: `ifu.sv`, `decompress.sv`, `spill.sv`, `irom.sv`, and `bpred/` (`bpred.sv`, `btb.sv`, `gshare.sv`, `gsharebasic.sv`, `icpred.sv`, `localaheadbp.sv`, `localbpbasic.sv`, `localrepairbp.sv`, `RASPredictor.sv`, `satCounter2.sv`, `twoBitPredictor.sv`). Where a fetch-side behavior is only observable through another unit, the relevant parts of `src/hazard/hazard.sv`, `src/ieu/controller.sv`, `src/ieu/alu.sv`, `src/privileged/{trap,csr,csrm,privpiperegs}.sv`, `src/mmu/{mmu,pmachecker,adrdecs}.sv`, `src/ebu/busfsm.sv`, and `src/cache/{cachefsm,cacheway}.sv` were read as well. Zicfilp is not implemented by Wally (no landing-pad logic exists), so `cfi.adoc` does not apply.
**Reference:** Ratified text in `~/repos/riscv-isa-manual/src/`: `unpriv/rv32.adoc` (JAL/JALR/branch semantics, instruction-address-misaligned), `unpriv/intro.adoc` (IALIGN/ILEN, encoding lengths, all-zero/all-ones illegal), `unpriv/zc.adoc`, `zca.adoc`, `zcb.adoc`, `zcd.adoc`, `zcf.adoc`, `c-st-ext.adoc` and the three quadrant listings `images/bytefield/rvc-instr-quad{0,1,2}.edn`; `unpriv/zifencei.adoc`; `priv/machine.adoc` (misa, mepc, mtval, exception priority, xRET, PMA idempotency); `priv/supervisor.adoc` (sepc, stval). Rule anchors (`norm:`) are cited where the text carries one.
**Baseline:** branch `cmpnv`, commit `1f56be750`.

## Summary

[...]

PR #1859, opened 2026-09-13
IFU: report a fetch fault on the first half of a spilled instruction (#1838 item 1)

davidharrishmc:
Issue #1838 item 1.

A fetch fault on the first half of a 32-bit instruction that spills across a cache line or page was discarded, because taking the spill stalls D and `faultregD` is the only register that samples the F-stage fault bits. When the first half faulted and the second did not, no exception was raised and the core executed a word merged from two bytes of the protected page.

`spill.sv` now latches the first half's fault status on the existing `SpillSaveF` enable and ORs it back while `SelSpillF` is asserted. `ifu.sv` clears the spill flag feeding `xtval` when the first half faulted, so `xtval` reports `PCM` rather than `PCM+2` — completing the cause-12/1 half of item 2, whose `ebreak` half landed in b787dd19a.

Suppressing `TakeSpillF` instead would be shorter, but it grafts the PMP priority decode onto the `StallF`/`D`/`E`/`M`/`W` network, roughly nine levels past where `CacheableF` already arrives. The commit message explains the tradeoff.

Independent of the other three PRs for this issue; they apply in any order.

**Verification:** `lint-wally` and `lint-wally --nightly` clean. The arch tests could not be rebuilt here (the installed `sail_riscv_sim` no longer accepts `--trace-all`), so all 44 `tests/coverage` ELFs were run on rv64gc against `main`: all 44 cycle-identical.

https://claude.ai/code/session_013H8QE6RS2J9nJu1x4sThUh

### DIFF
--- a/src/ifu/ifu.sv
+++ b/src/ifu/ifu.sv
@@ -142,6 +142,7 @@ module ifu import cvw::*;  #(parameter cvw_t P) (
   logic [31:0]                 ShiftUncachedInstr;
   logic                        ITLBMissF;
   logic                        InstrUpdateAF;                            // ITLB hit needs to update dirty or access bits
+  logic                        InstrPageFaultRawF, InstrAccessFaultRawF; // Faults on the half of the fetch currently being translated
 
   assign PCFExt = {2'b00, PCSpillF};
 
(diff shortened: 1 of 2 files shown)


rosethompson:
I've reviewed the PR and I agree with the changes.  There is now a merge conflict. Can you resolve?

PR #1860, opened 2026-09-13
IFU: do not fetch or cache an instruction that failed the PMA or PMP check (#1838 item 3)

davidharrishmc:
Issue #1838 item 3.

An instruction fetch into a peripheral region raised the access fault correctly but still issued the AHB read first, so the device saw it: reading the PLIC claim register claims the highest-priority pending interrupt, and reading the UART RBR pops the receive FIFO. The spec forbids early or speculative implicit accesses to non-idempotent regions, and the access is redundant by construction since the PMA marks peripherals RW, not RX.

The two fetch faults are folded into `IFUFaultF`, which gates both `BusRW` assignments and `CacheRWF`. With `BusRW` low the bus FSM stays in `ADR_PHASE` with `HTRANS` idle, `BusStall` low and `BusCommitted` low, so nothing hangs and the trap is not deferred.

### Why the cache is gated too

Gating only `BusRW` left the cacheable path untouched, so a fetch that failed the PMA or PMP check still ran a cache access and, on a miss, still filled the line over the bus. That is never architecturally visible — the trap in the Memory stage wins and the data is never returned — but the denied line was resident in the I$ afterward, which is a needless cache side channel and contradicts the intent of the fix.

`CacheRWF` is the right place rather than `CacheBusRW`. The cache FSM derives `AnyMiss`, `CacheBusRW`, `CacheStall`, and its next state from `CacheRW`, so suppressing only the bus request would leave the FSM stalling in `STATE_ACCESS` on a miss it can never fill. One gate on `CacheRWF` keeps all four consistent.

A faulting fetch now stops stalling for a line fill and goes straight to the trap, matching what the uncached path already did once `BusRW` was gated.

### Timing

The I$ is VIPT — the index comes from `PCSpillNextF[11:0]` a cycle early, the tag from `PCPF` — so in the critical cycle the PMA/PMP check and the tag compare are parallel branches off the same ITLB output rather than a serial addition. The added delay is the excess of the fault path over the tag compare, plus one AND. Worth a `synthDC` check on rv64gc before merge, though the LSU path is usually the critical one in Wally.

Independent of the other three PRs for this issue; they apply in any order.

### Verification

`lint-wally` and `lint-wally --nightly` clean. `verilator --lint-only` is also clean on rv32e, rv64gc, rv32gc, rv32imc, rv32i and rv64i, which covers the MMU/no-MMU, icache/passthrough and bus/no-bus generate branches that `IFUFaultF` crosses.

The arch tests could not be rebuilt here (the installed `sail_riscv_sim` no longer accepts `--trace-all`), so the `coverage64gc` suite (39 tests) was run on rv64gc under Verilator from two clean worktrees, one at `main` (2f82f0011) and one at this branch. Both report `SUCCESS! All tests ran without failures.`

To get per-test timing the coverage `$display` was temporarily given a `$time` stamp, identically in both trees and not committed. The testbench clock is 10 time units, so 10 ps of `$time` is one cycle.

Suite total: **940,117 cycles on `main`, 939,146 on this branch — 971 cycles faster (0.10%)**. 19 of the 39 tests changed; 18 got faster and one got slower:

| test | delta (cycles) |
| --- | --- |
| tlbNAPOT | -420 |
| tlbGP | -128 |
| tlbASIDMISS | -82 |
| tlbmisc | -73 |
| vm64check | -71 |
| tlbGLBHIT | -60 |
| tlbGLB, btbthrash | -30 each |
| dcache1 | -18 |
| pmp | -14 |
| tlbASID, pmpadrdecs | -10 each |
| pmppriority | -8 |
| hptwAccessFault, nonleafpbmtfault | -6 each |
| tlbTP, dcache2 | -4 each |
| floatmisc | -2 |
| **tlbGLBASID** | **+5** |

Every test that moved is a TLB, VM or PMP test that fetches from a faulting address, which is what the mechanism predicts: those fetches no longer wait for a line fill they are not allowed to use. `tlbGLBASID` being 5 cycles slower is expected rather than a regression — once the denied line is no longer filled, a later legitimate fetch to that line misses where it previously hit on the speculatively cached copy.

https://claude.ai/code/session_01JjSeYms8a7Ds1YRiyXKNzW

### DIFF
--- a/src/ifu/ifu.sv
+++ b/src/ifu/ifu.sv
@@ -142,6 +142,7 @@ module ifu import cvw::*;  #(parameter cvw_t P) (
   logic [31:0]                 ShiftUncachedInstr;
   logic                        ITLBMissF;
   logic                        InstrUpdateAF;                            // ITLB hit needs to update dirty or access bits
+  logic                        IFUFaultF;                                // Fetch failed the PMA or PMP check, so it must not access the cache or bus
 
   assign PCFExt = {2'b00, PCSpillF};
 
@@ -214,6 +215,11 @@ module ifu import cvw::*;  #(parameter cvw_t P) (
     assign ITLBMissOrUpdateAF = '0;
   end
 
+  // A fetch that has already failed the PMA or PMP check traps in the Memory stage, so it must not
+  // start a bus transfer or a cache line fill.  Gating the cache as well as the bus keeps a line
+  // that the hart may not execute from being filled into the I$.
+  assign IFUFaultF = InstrAccessFaultF | InstrPageFaultF;
+
   ////////////////////////////////////////////////////////////////////////////////////////////////
   // Memory
   ////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


rosethompson:
This also should fix @juanschroeder issue with the BTB sending out speculated fetches posted in issue #1768.  It does not fix issue #1768.

PR #1863, opened 2026-09-13
Fix Zicbom instruction decoding when extension is unsupportedFix Zicbom instruction decoding

Haiqua-Ghaffar:
Fixes an incorrect CMOFunctD expression in src/ieu/controller.sv caused by operator-precedence/grouping.

The cbo.clean and cbo.flush decode conditions were not properly gated by P.ZICBOM_SUPPORTED. As a result, these instructions could be decoded as legal CMO instructions even when the Zicbom extension was disabled, instead of raising an Illegal Instruction exception.

The expression has been corrected so that all Zicbom instructions (cbo.inval, cbo.clean, and cbo.flush) are properly gated by P.ZICBOM_SUPPORTED.

This is a minimal RTL-only fix that changes only the grouping of the existing logic.

All regression tests continue to pass.


davidharrishmc:
Fixed in #1857 a few hours earlier.

PR #1864, opened 2026-09-13
Fix illegal mcounteren access when U-mode is unsupported

Syed-Moeed-Ali:
Fixes an illegal-instruction exception issue for MCOUNTEREN accesses when P.U_SUPPORTED=0.

According to the RISC-V Privileged Architecture Specification, mcounteren does not exist when U-mode is unsupported, and attempts to access it must raise an Illegal Instruction exception.

Previously, the MCOUNTEREN read path in src/privileged/csrm.sv unconditionally returned the register value without checking P.U_SUPPORTED, allowing CSR reads to proceed without trapping.

The access is now gated by P.U_SUPPORTED, following the existing pattern used by adjacent CSRs such as MENVCFG.


davidharrishmc (approved):
Thank you for this fix.

davidharrishmc:
All contributors need to be covered by a signed Eclipse Contributor Agreement indicating this work becomes open source. 

You need to create an account with https://www.eclipse.org/
Sign the agreement, https://www.eclipse.org/legal/ECA.php

Thank you,

David Harris

PR #1865, opened 2026-09-13
wsim: report a nonzero exit status when the build or the simulation fails

davidharrishmc:
`wsim` exited 0 no matter what happened.

A Verilator, Questa or VCS compilation failure hit `if os.system(compile_cmd): return`, which returns `None` from the per-simulator function. `runSim` then discarded that return value and returned `None` itself, so `sys.exit(runSim(...))` always exited 0. I found this while running a sweep of ELFs against a tree whose `verilog-ethernet` submodule was not populated: every test reported `rc=0` while not one of them had compiled.

The simulation phase was worse — its status was never examined at all for any of the three simulators, so a simulator crash also exited 0.

### Changes

All in `bin/wsim`; `main()` is untouched, since `sys.exit(runSim(...))` was already correct and was simply being handed `None`.

- `runSim` propagates each per-simulator return value instead of dropping it, and reports an unknown simulator rather than falling off the end.
- Compilation failures print which simulator and configuration failed, on stderr.
- Every command now goes through `subprocess.run(cmd, shell=True).returncode`, which is the exit code directly. This matters: `os.system` returns a wait *status*, and handing that to `sys.exit` turns an exit code of 1 into 256, which exits 0 — the naive fix would have reproduced the bug for the most common failure code.
- `checkCompiled` covers the case where compilation was *skipped* because another process held the compile lock, but that process failed to leave a design behind.

### Testing

- Build failure (missing submodule): now exits 2 with `Error: Verilator compilation failed for rv64gc testbench`. Previously exited 0 in silence.
- Success path: `wsim rv64gc coverage64gc --sim verilator`, 39/39 tests pass, exits 0.
- `ruff check bin/wsim` clean.
- Verified the exit codes directly: `os.system('exit 1')` returns 256 and `sys.exit(256)` exits 0, whereas `subprocess.run('exit 1', shell=True).returncode` is 1.

`regression-wally` is unaffected in its verdicts because it greps logs for a success string. It does check the command status first, so it will now report `Execution failed` rather than the vaguer `No success string found` when a build breaks.

### Related cases deliberately left alone

Two adjacent gaps that this PR does not touch, because each needs a judgement call rather than a mechanical fix:

1. **`testbench.sv:290` reports hard errors with `$finish`, so they still exit 0.** `wsim rv64gc nosuchtestsuite` prints `TEST nosuchtestsuite not supported in this configuration` and exits 0 even with this change, because the testbench itself returns 0. Lines 170 and 173 have the same shape. Switching them to `$fatal` would fix it but touches the `$stop`/`$finish` behaviour that the comments at lines 488 and 508 say Questa's `wally-batch.do` depends on.
2. **`sim/Makefile:98-99` redirect rather than pipe,** so those `--fcov` targets will now fail the make target if `wsim` fails. No edit needed, but it may surface something that was previously silent.

### Checked and already correct

`.github/workflows/install.yml:220` pipes `wsim` through `tee`, which would normally discard the exit status. It does not here: the workflow sets `defaults: run: shell: bash`, and an explicit `shell: bash` in GitHub Actions runs `bash --noprofile --norc -eo pipefail {0}`, so `pipefail` is already in effect and a `wsim` failure will fail that step.

https://claude.ai/code/session_01JjSeYms8a7Ds1YRiyXKNzW


bin/wsim
```diff
@@ -196,13 +196,32 @@ def targetMtime(target):
     return target.stat().st_mtime if target.exists() else None
 
 
+def runCmd(cmd):
```
jordancarlin: Seems like switching to subproccess.run instead of os.system would be cleaner and avoid the need for this helper function.

davidharrishmc: done

Issue #1802, opened 2026-07-15
cvw should only count hpmevents when mhpmevent is nonzero


davidharrishmc:

davidharrishmc:
Closed with #1866

PR #1866, opened 2026-09-14
Counter Redesign

davidharrishmc:
Fixed multiple issues in counters:
* Added writable mphmevent CSRs
* Counters only increment when mhpmevent is nonzero #1802
* Fixes upper mphmcounter/hpmcounter  being read-only-zero instead of faulting
* Bits in m/scoutneren and mcountinhibit are roz if counter is read-only zero or inaccessible from lower modes
* Add derived configurations to exercise these cases


src/privileged/csrm.sv
```diff
@@ -104,6 +104,10 @@ module csrm  import cvw::*;  #(parameter cvw_t P) (
   // when compressed instructions are supported, there can't be misaligned instructions
   localparam MEDELEG_MASK  = P.ZCA_SUPPORTED ? 16'hB3FE : 16'hB3FF;
   localparam MIDELEG_MASK  = 12'h222; // we choose to not make machine interrupts delegable
+  localparam COUNTERINHIBIT_MASK = ((1 << P.COUNTERS) - 1); // only nonzero counters can be inhibited
```
rosethompson, line 107: Can counter 1 be inhibited?

src/privileged/csrc.sv
```diff
@@ -122,73 +131,107 @@ module csrc  import cvw::*;  #(parameter cvw_t P) (
...
+      end
+    else if (PrivilegeModeW == P.M_MODE |
+      MCOUNTEREN_REGW[CounterNumM] & (!P.S_SUPPORTED | PrivilegeModeW == P.S_MODE | SCOUNTEREN_REGW[CounterNumM])) begin
+        // The branch conditions below guarantee CounterNumM < P.COUNTERS before it indexes the
+        // counter arrays, but Verilator sizes the index from the array bound, so it warns whenever
+        // P.COUNTERS < 32.  This region also covers the MTIME_CLINT reads, which Verilator does not
+        // realize happen only at one XLEN.
+        /* verilator lint_off WIDTH */
         if (P.XLEN==64) begin // 64-bit counter reads
           // Veri lator doesn't realize this only occurs for XLEN=64
```
rosethompson, line 194: Typo.

davidharrishmc: Veri lator is because // Verilator is a special directive.

src/privileged/csrm.sv
davidharrishmc: Good catch

PR #1867, opened 2026-09-14
Fix/sfenceFix sfence.vma decoding when virtual memory is unsupported vma no virtmem

Haiqua-Ghaffar:
Fixes a spurious Illegal Instruction exception on sfence.vma when S-mode is supported but virtual memory is disabled.

In src/privileged/privdec.sv, the decode logic for sfencevmaM was incorrectly gated by P.VIRTMEM_SUPPORTED instead of P.S_SUPPORTED.

As a result, in configurations such as nodcache_rv64gc, where S-mode is supported (P.S_SUPPORTED=1) but virtual memory is not (P.VIRTMEM_SUPPORTED=0), sfence.vma was not decoded as a valid privileged instruction. This caused IllegalPrivilegedInstrM to assert and generated an Illegal Instruction exception (Cause 2).

According to the RISC-V Privileged Architecture Specification, SFENCE.VMA is valid when S-mode is implemented. When address translation is not supported and satp.MODE is Bare, the instruction executes as a no-op rather than raising an exception.

The decode condition has been corrected to gate sfencevmaM with P.S_SUPPORTED


davidharrishmc:
Priv spec says 
```
For implementations that make satp.MODE read-only zero (always Bare), attempts to execute an
SFENCE.VMA instruction might raise an illegal-instruction exception.
```
so cvw behavior conforms with spec.

PR #1869, opened 2026-09-14
Fix cbo.inval decode when CBIE=01

abdulrafay7038:
**Bug:**
When ENVCFG_CBE[1:0] == 2'b01, cbo.inval was incorrectly decoded as cbo.clean in src/ieu/controller.sv.

The CBIE=01 condition was OR'd into CMOpD[1], while CMOpD[2] is the existing decode path for cbo.flush. As a result, cbo.inval with CBIE=01 followed the clean operation path and did not invalidate the cache line.

**Fix:**

Move the CBIE=01 cbo.inval condition from CMOpD[1] to CMOpD[2].

This causes the operation to use the existing flush path, which asserts ClearValid and correctly invalidates the cache line.


davidharrishmc:
Fixed with #1857.

PR #1870, opened 2026-09-15
Fix illegal mstatus.MPP reset value when U-mode is unsupported

Haiqua-Ghaffar:
Fixes an illegal reset value of mstatus.MPP in configurations where User mode is not supported (e.g. noU_rv64gc, where P.U_SUPPORTED=0 and P.S_SUPPORTED=0).

According to the RISC-V Privileged Architecture Specification, the xPP fields are WARL fields that can hold only privilege mode x and any implemented privilege mode lower than x. Since M-mode is always implemented, this means that when neither S-mode nor U-mode is supported, MPP can only ever legally hold 2'b11 (M-mode).

Previously, src/privileged/csrsr.sv unconditionally reset STATUS_MPP to 2'b00 (User mode) on hardware reset, regardless of P.U_SUPPORTED. In configurations without U-mode or S-mode support, this leaves MPP in an illegal, unsupported state immediately after reset. Because the WARL write logic already correctly rejects writes of 2'b00/2'b01 when those modes aren't supported, there was no way for software to correct the value — MPP remained stuck at the illegal reset value indefinitely.

The fix mirrors the existing, correct logic already used in the mretM path a few lines below, which sets MPP to the lowest supported privilege level:

systemverilog
STATUS_MPP <= P.U_SUPPORTED ? P.U_MODE : P.M_MODE;

Applying the same expression to the reset assignment ensures MPP always resets to a legal value: User mode when supported, otherwise Machine mode.

Files changed: src/privileged/csrsr.sv (single-line change)


davidharrishmc:
Thank you for the PR.  May I ask what led you to investigate this?

Haiqua-Ghaffar:
Thank you for asking. I am an Electrical Engineering student at UET Lahore, Pakistan and part of the HiCoar Research Group. We are currently participating in Hackathon, where we are exploring bug discovery and verification in Wally. This is our group's third accepted PR arising from that work.

davidharrishmc:
Thank you for contributing.

PR #1871, opened 2026-09-15
Possible fix for #1538/1766

davidharrishmc:


rosethompson:
Good news! The FPGA still works after this fix. I still want to do a very careful review.

PR #1872, opened 2026-09-15
mmu: preserve global TLB entries on an ASID-scoped sfence.vma

davidharrishmc:
Splits the `sfence.vma` half out of the reverted PR #1768 (issue #1538) so it can be reviewed on its own. The hardware page table walker, the LSU arbitration and the caches are untouched here; only the TLB flush path changes.

## What changes

The privileged spec requires `sfence.vma` with `rs2 != x0` to invalidate translations only for the address space named by `rs2`, *"except for entries containing global mappings"*. Global (G=1) entries must be invalidated only when `rs2 = x0`. Wally invalidates every TLB entry on any `sfence.vma`.

`privdec` now decodes

```systemverilog
assign sfencevmaAllM = sfencevmaM & ~|InstrM[24:20];   // rs2 == x0
```

which is plumbed as `TLBFlushAll` through `privileged` → `ifu`/`lsu` → `mmu` → `tlb` → `tlbcam` → `tlbcamline`, where the unconditional flush becomes

```systemverilog
assign ShouldFlush = TLBFlush & (~PTE_G | TLBFlushAll);
flopenr #(1) validbitflop(clk, reset, WriteEnable | ShouldFlush, ~ShouldFlush, Valid);
```

Nine RTL files, 32 insertions.

## Why, and what this is not

The old behaviour was **not** a conformance bug. The spec explicitly permits an implementation to invalidate more entries than required, so flushing everything was legal. What it costs is that every ASID-scoped fence throws away the global kernel mappings, which makes tagging an entry global pointless and forces a full ITLB/DTLB refill on each address-space switch.

So the risk this introduces runs the other way: a global entry that *should* be flushed but is not. That is what the new test guards.

Coverage of the four `rs1`/`rs2` combinations after this change:

| rs1 | rs2 | Spec requires | This implementation |
|---|---|---|---|
| x0 | x0 | invalidate everything, globals included | flushes all entries |
| != x0 | x0 | that address, all address spaces, globals included | flushes all entries (ignores rs1, permitted) |
| x0 | != x0 | that ASID, globals preserved | flushes all non-global entries (ignores the ASID value, permitted) |
| != x0 | != x0 | that address in that ASID, globals preserved | flushes all non-global entries (permitted) |

`sinval.vma` follows the same rs2 rule. `sfence.w.inval` and `sfence.inval.ir` have rs2 = x0 encoded, so they still flush everything.

## Test

`tests/coverage/sfencevmaglobal.S` (added to the `coverage64gc` list) is self-checking through its own `selfcheck_record` and builds its Sv48 tables at runtime. It maps one VA to either of two data pages and another VA to either of two code pages, then checks that:

1. a global mapping is invalidated by `sfence.vma x0, x0`, on the data TLB;
2. it is also invalidated by the address-scoped `sfence.vma rs1, x0`;
3. an ASID-scoped `sfence.vma x0, rs2` still invalidates a **non**-global mapping;
4. a global mapping still translates correctly across an ASID-scoped fence, whether the entry is preserved or refilled;
5. the same required-invalidation cases on the **instruction** TLB, by calling into a page whose mapping is switched between two code pages.

It passes on unmodified `main` as well, because it tests the direction the spec requires and the old code satisfies that by flushing everything. To confirm it is not vacuous, I inverted the new condition to `ShouldFlush = TLBFlush & ~PTE_G` (never flush globals) and it fails on check 2, reading the stale page:

```
Error on test sfencevmaglobal.elf result 2: sim (D$) 000000000a0a0a0a signature = 000000000b0b0b0b
```

## Verification

| Check | Result |
|---|---|
| `lint-wally` (rv32e, rv64gc, rv32gc, rv32imc, rv32i, rv64i, fdqh_rv64gc) | clean |
| `regression-wally` (Verilator) | 75 suites pass |
| New test, Verilator + ImperasDV lockstep | passes, 0 mismatches |
| `tlbGLB`, `tlbGLBASID`, `tlbGLBHIT`, `tlbASID`, `tlbASIDMISS`, `tlbmisc`, `tlbNAPOT` lockstep | 0 mismatches each |
| Full buildroot Linux boot | boots to a root shell, no faults |

`arch64vm_sv57` could not be run: the pinned `riscv-arch-test` submodule has no `vm_sv57` directory in my environment. It fails identically on `main`.

Because PR #1768 was reverted for an FPGA Linux boot failure, I ran a full buildroot boot in simulation on this branch specifically to exercise that half of the PR in isolation. Wally implements 16 ASID bits, so Linux uses ASID-scoped fences on context switch and this change is live throughout userspace.

The boot runs to completion in about 550M instructions, straight through the point where the FPGA failure was reported:

```
[   27.267273] Freeing initrd memory: 71680K
[   29.361429] Freeing unused kernel image (initmem) memory: 73816K
[   29.371921] Run /init as init process
init started: BusyBox v1.36.1
Starting syslogd: OK / Starting klogd: OK / Running sysctl: OK / Starting crond: OK
...
Hello this ~/.profile is meant to demonstrate running some basic commands on Wally.
I am root
And I am on WallyHostname
```

The revert log shows `BUG: Bad page state in process kworker/u4:0` at 29.384866. This boot passes 29.37 and reaches an interactive root shell with no bad page state, oops, panic, call trace or segfault anywhere in the console log. A boot on unmodified `main`, run side by side, produced byte-identical console output over the same interval.

Worth stating plainly: this is simulation, and the FPGA report involved a kworker faulting while handling `libstdc++.so.6.0.33`, which this initramfs workload never loads. It is strong evidence rather than proof for the FPGA, so a hardware boot is still worth doing. (The `/sbin/ifup` and crontab complaints during the boot are missing files in the minimal image and appear on `main` too.)

## Review note

One corner case, unchanged in spirit from the existing design: if a page table walk writes a TLB line in the same cycle an ASID-scoped fence pulses, `PTE_G` is still the *previous* contents' global bit, so a line that previously held a global entry is protected from that fence while being overwritten. The entry that results was just built from fresh page table state, so it is not stale; the pre-existing `CAMHit = |Matches & ~TLBFlush` term covers the same cycle.


davidharrishmc:
The buildroot boot has now run to completion, so the evidence above can be stated more strongly than when I opened this.

The boot reaches `WallyHostname login:`, the string `regression-wally` greps for on the full-boot test, after 576M instructions, and ends through the documented normal path (`watchdog.sv:80`: *"Watch Dog Time Out triggered. This is a normal termination for a full buildroot boot"*), not a stuck-PC or recursive-trap failure.

Before the prompt, the demo profile exercises real userspace work with no faults:

```
Created myFile.txt and moved it to myDir. It contains: Hello World! / And farewell!
Created myScript.sh. Running it yields: Hello this is another example script
Created symLinkToMyScript.sh. Running it yields: Hello this is another example script
Here is disk usage: devtmpfs 48.8M / tmpfs 119.8M ...
And here are the current processes:
    1 root      2160 S    init
    8 root         0 IW   [kworker/u4:0-as]
   58 root      2160 S    -/bin/sh
And finally a login prompt.
WallyHostname login:
```

Note `kworker/u4:0` is alive and scheduling: that is the exact process named in the revert's `BUG: Bad page state in process kworker/u4:0`. Across the whole 10379-byte console log there is no bad page state, oops, panic, call trace or segfault.

The FPGA caveat in the description still stands: this workload never loads `libstdc++.so.6.0.33`, so a hardware boot is still worth doing before this is trusted on the board.

rosethompson:
This merge looks good except for the conflict.

PR #1874, opened 2026-09-15
Clear reported dependency vulnerabilities in uv.lock

davidharrishmc:
Clears 11 of the 13 GitHub security advisories reported against `uv.lock`, covering filelock, Pillow, requests, pytest, urllib3, cryptography, setuptools, GitPython, and mistune.

The two PyYAML advisories (CVE-2020-1747, CVE-2020-14343) are **knowingly left unfixed**. `riscv-config` hard-pins `pyyaml==5.2` in its `setup.py`, still true on its upstream master and latest release. Overriding that pin risks unexpected behavior in riscof, so per review we live with it until RISCOF is retired.

## Root cause

The lock had a forked resolution, one branch for Python 3.9 and one for newer interpreters. Nearly every vulnerable version existed **only** in the 3.9 branch, because the patched releases no longer build for 3.9. Pinning versions could not fix this; the floor had to move.

## Changes

One hand edit to `pyproject.toml`: **`requires-python` moves from 3.9 to 3.10.** That collapses the old fork and picks up the patched versions. `uv.lock` is regenerated by `uv lock --upgrade`. The lock drops from 174 to 137 package entries and from five resolution markers to three.

| package | was | now | patched in |
|---|---|---|---|
| PyYAML | 5.2 | 5.2 (unfixed, see above) | 5.4 |
| filelock | 3.19.1 | 3.32.6 | 3.20.1 |
| Pillow | 11.3.0 | 12.3.0 | 12.3.0 |
| requests | 2.32.5 | 2.34.2 | 2.33.0 |
| pytest | 8.4.2 | 9.1.1 | 9.0.3 |
| urllib3 | 2.6.3 | 2.8.0 | 2.7.0 |
| cryptography | 47.0.0 | 50.0.1 | 48.0.1 |
| setuptools | 82.0.1 | 84.0.0 | 83.0.0 |
| GitPython | 3.1.50 | 3.1.62 | 3.1.52 |
| mistune | 3.3.2 | 3.3.4 | 3.3.3 |

Every one of these is now a single version in the lock rather than forked, so any supported interpreter gets the patched release.

## Older distro compatibility

No regressions, and one improvement.

- **Only two supported distros are affected.** `wally-package-install.sh` already installs Python 3.11 or newer nearly everywhere: 3.12 on all RHEL versions, 3.14 on Ubuntu 26.04, 3.12 on 24.04, 3.11 on 22.04, 3.13 on Debian 13, 3.11 on Debian 12, 3.13 on SUSE 16.0, 3.12 on SUSE 15.6. The only two below the new floor are **Ubuntu 20.04 and Debian 11**, which install 3.9. On those, uv downloads a managed interpreter for the project venv instead of using the installed 3.9. It does this automatically; there is no `.python-version`, no `UV_PYTHON_DOWNLOADS` setting, and no mise Python pin to block it. Verified on a RHEL 9 host whose `python3` is 3.9.25: `uv sync` silently fetched a managed interpreter and installed cleanly.
- The `Lint (Python 39)` CI job is unaffected. It only rewrites ruff's `target-version` and runs pre-commit; it never installs the project. Both lint jobs pass on this PR.
- **Lock format unchanged** at `version = 1`, `revision = 3`, so the uv 0.11.0 pinned by `uv-install.sh` reads it fine.
- **glibc floors unchanged.** The binding constraint was already lief at glibc 2.28, which RHEL 8 meets exactly, and lief still requires 2.28. Only Pillow tightened, 2.17 to 2.27, still below RHEL 8's 2.28.
- **Source builds unchanged.** PyYAML 5.2 ships no wheels, so uv builds it from the sdist as before; it builds cleanly for the interpreters in scope, falling back to its pure-Python implementation when libyaml headers are absent. python-constraint remains the only other source-only package and is pure Python.

## Testing

Re-verified after dropping the override, with PyYAML back at 5.2:

- `riscof validateyaml` runs clean against the rv64gc config.
- `riscof testlist` builds the full database and selects 1835 tests.
- uv builds PyYAML 5.2 from sdist for the selected interpreter without error.
- `pre-commit` passes, and every direct dependency imports.
- Both plotting scripts compile, and `wsim` is unaffected since it runs its own uv script environment.

## Note for reviewers

`riscv-config` and `riscv-isac` are reached only through riscof; nothing in cvw imports either directly, and riscv-isac is never invoked. When riscof is retired, dropping all three takes the lock from 137 packages to 40, removes eight of the ten packages in the table above outright, and makes the PyYAML override unnecessary. Worth doing as a follow-up rather than here.


pyproject.toml
```diff
@@ -18,3 +18,8 @@ dependencies = [
   "riscv-config @ git+https://github.com/riscv-software-src/riscv-config.git@54171f205be802f9f8e0b1cf4156a6cc826fb467",                     # May 31, 2025
   "riscv-isac @ git+https://github.com/riscv-non-isa/riscv-arch-test.git@450de2eabfe4fcdfdf54135b5ab2dbb1d94805f8#subdirectory=riscv-isac", # May 31, 2025
 ]
+
+[tool.uv]
+# riscv-config hard-pins pyyaml==5.2 (CVE-2020-1747, CVE-2020-14343) in its setup.py,
+# still true on upstream master, so override the pin to get a patched PyYAML.
+override-dependencies = ["pyyaml>=6.0.3"]
```
jordancarlin: Revert this change. We will need to live with the PyYAML vulnerability until we drop RISCOF. This will lead to unexpected results.

davidharrishmc: Done, reverted in 346a72b40. PyYAML is back to 5.2 and the `[tool.uv]` block is gone.

Re-verified with the pin restored: uv builds PyYAML 5.2 from its sdist for the selected interpreter without error, `riscof validateyaml` passes on the rv64gc config, and `riscof testlist` still selects 1835 tests.

The other 11 advisories stay fixed, since those came from collapsing the Python 3.9 resolution branch rather than from the override. I updated the PR description to record the two PyYAML CVEs as knowingly accepted until RISCOF is dropped.

jordancarlin:
Looks fine now.

PR #1875, opened 2026-09-15
This PR depends on PR #1871 and simplifies the bug in that PR

rosethompson:
The AI made some assumptions which are incorrect.  It assumes the CPU's M-stage memory access can occur between a ITLB or DTLB miss.  This should not happen by construction. The correct sequence of events are to handle the DTLB miss, the ITLB miss, and then the CPU's memory access.  I suspect the possibility of an ITLB miss with a first memory access hitting the D$ might confuse the AI.  However, the purpose of the hptw.sv fixes were to enforce this behavior.
    
Regardless the AI did fix the underlying problems.  The first was the enforcement of the sequence DTLB miss, ITLB miss, and then memory access. Second was SelHPTWAdr needed to be lowered after a fault so that a memory request could be reissued.  The specific error in issue #1538 is ITLB miss causes a fault followed by a load/store hit or miss.  Because the HPTW finished in the FAULT state it did not lower SelHPTWAdr and the next cycle the cache had the wrong tag for the load/store.
    
This commit removes all of the ReadDataM holding that should not be required if the above requirement is implemented correctly.

Must merge PR #1871 first.


davidharrishmc (approved):
Please delete the commented-out lines rather than leaving them stale.  Also there is a long Claude comment followed by a long Rose Thompson comment.  Strip out the history and just leave the final conclusion.

Otherwise good to merge.

PR #1876, opened 2026-09-16
Add Zimop / Zcmop support

davidharrishmc:


src/ieu/controller.sv
```diff
@@ -216,6 +217,9 @@ module controller import cvw::*;  #(parameter cvw_t P) (
     assign JRFunctD         = Funct3D == 3'b000;
     assign PFunctD          = Funct3D == 3'b000 & RdD == 5'b0;
     assign CSRFunctD        = Funct3D[1:0] != 2'b00;
+    // Zimop: mop.r.n is 1_n[4]_00_n[3:2]_0111_n[1:0], mop.rr.n is 1_n[2]_00_n[1:0]_1_rs2, both with funct3 = 100
+    assign MOPFunctD        = (Funct3D == 3'b100) & InstrD[31] & (InstrD[29:28] == 2'b00) &
+                              (InstrD[25] | (InstrD[24:22] == 3'b111));
```
rosethompson, line 222: Should (InstrD[24:22] == 3'b111) be (InstrD[25:22] == 4'b0111)?

src/ifu/decompress.sv
rosethompson, line 1: Should c.mop instructions have immCILUI equal to 0 not != 0?

rosethompson (dismissed):

src/ieu/controller.sv
davidharrishmc: If D[25] = 1, then it is a a MOPRR rather than MOPR, so it is a MOP in either case.

src/ifu/decompress.sv
davidharrishmc: the imm != 0 case captures c.lui.  otherwise we go to the else and capture c.mop.  so c.mop has imm = 0.

rosethompson: Ahh. That is correct.

PR #1877, opened 2026-09-16
Zabha support

davidharrishmc:


src/lsu/amoalu.sv
```diff
@@ -37,28 +37,52 @@ module amoalu import cvw::*;  #(parameter cvw_t P) (
...
+    comparator #(32) cmpw1(a[63:32], b[63:32], 1'b0, {eqW1, ltW1});
+    assign ltu64 = ltW1 | (eqW1 & ltu32);
+    assign lt64  = (sngd & (a[63] ^ b[63])) ? a[63] : ltu64;
   end
 
+  // Pick the compare matching the access width
+  always_comb
+    case (LSUFunct3M[1:0])
+      2'b00:   lt = P.ZABHA_SUPPORTED ? lt8  : lt32; // amo*.b
+      2'b01:   lt = P.ZABHA_SUPPORTED ? lt16 : lt32; // amo*.h
```
rosethompson, line 82: Is selecting lt32 for no ZABHA a synthesis optimization? If so we should comment.

davidharrishmc: Yes, purely a synthesis optimization, and you are right that it wanted a comment. Added in f9693726f.

Without Zabha, `AFunctD` in controller.sv rejects funct3 = 000 and 001, so a byte or halfword AMO is an illegal encoding that never reaches the LSU and `lt` is a don't-care in those two arms. Writing `lt8`/`lt16` there would leave them as real consumers of the narrow signed-compare muxes, which the tool cannot prove unreachable; selecting `lt32` instead leaves `lt8` and `lt16` with no consumer at all, so those two muxes and their sign XORs are swept. The lane comparators themselves stay either way, since `ltu16` still feeds `ltu32`.

The comment now reads:

```systemverilog
  // Pick the compare matching the access width.  Without Zabha, byte and halfword AMOs are illegal
  // encodings that never reach the LSU (see AFunctD in controller.sv), so those two cases are
  // don't-cares.  Selecting lt32 for them is a synthesis optimization: it leaves lt8 and lt16 with no
  // consumer, so the narrow signed-compare muxes are trimmed away.
```

The branch is also rebased onto current main to clear the conflicts (Zimop/Zcmop landed next to the Zabha lines in both `imperas.ic` files and in `riscvassertions.sv`; both sides kept).

davidharrishmc:
Ran the ACT4 tests against this branch with `make EXTENSIONS=Zabha,Zaamo cvw`, since nothing in the cvw regression exercises byte or halfword AMOs.

| Config | Result |
|---|---|
| cvw-rv64gc | 18/18 Zabha pass |
| cvw-rv32gc | 27/27 pass (18 Zabha + 9 Zaamo) |
| cvw-rv64gc-counters3 | 18/18 pass |
| cvw-rv64gc-counters5 | 18/18 pass |
| cvw-rv64gc-counters5-nohpm | 18/18 pass |
| cvw-rv32gc-counters5-nocntr | 9/9 pass (Zaamo only; its UDB config does not list Zabha) |
| cvw-rv32imc | no tests, as expected without A |

The Zabha set is `amoadd`, `amoand`, `amoor`, `amoxor`, `amoswap`, `amomin`, `amomax`, `amominu`, `amomaxu`, each at `.b` and `.h`. That covers every narrow path in the comparator rework, signed and unsigned, which is what I most wanted checked given `lt8`/`lt16` and the lane assembly are new.

Two notes on the run, neither a problem with this branch:

- The rv64gc run also reported 5 failures, all `Zacas-*` and `ZacasZabha-*` ELFs left over in the work directory from the Zacas branch. This branch does not implement `amocas`, so those are expected.
- The first `make ... cvw` pass exited 2 because two configs had no ELFs built yet (`cvw-rv32imc`, which legitimately has no atomics, and `cvw-rv32gc`). Re-running `make EXTENSIONS=Zabha,Zaamo cvw-rv32gc` built 54 ELFs and passed 27/27.

Also rebased onto current main and resolved the conflicts (Zimop/Zcmop landed adjacent to the Zabha lines in both `imperas.ic` files and in `riscvassertions.sv`; both sides kept). `lint-wally` is clean, and `arch64a_amo`, `wally64a_lrsc`, `arch32a_amo`, `wally32a_lrsc`, `coverage64gc`, `arch64vm_sv48`, `arch32vm_sv32`, `wally32priv` and the periph suites all pass.

rosethompson (approved):

PR #1882, opened 2026-09-16
Fix off-by-three address decode on mhpmevent CSR writes

abdulrafay7038:
## Summary

`mhpmevent3`–`mhpmevent31` occupy CSR addresses `0x323`–`0x33F`. `src/privileged/csrc.sv` decodes writes to them as:

```systemverilog
localparam MHPMEVENTBASE = 12'h323;
...
for (i = 3; i < P.COUNTERS; i = i+1) begin : mhpmevent
  assign WriteMHPMEVENTM[i] = CSRMWriteM & (CSRAdrM == MHPMEVENTBASE + i);
```

`MHPMEVENTBASE` is already the address of `mhpmevent3`, and `i` is the absolute counter number starting at 3, so `MHPMEVENTBASE + i` overshoots by three. Index 3 decodes to `0x326`, index 4 to `0x327`, and so on.

## Problem

The read path indexes by the absolute counter number instead (`CounterNumM = CSRAdrM[4:0]`, so address `0x323` reads `MHPMEVENT_REGW[3]`), so reads and writes disagree about which register a given address names. On `rv64gc` this produces:

- `csrw mhpmevent3` (`0x323`) asserts no write enable at all; the register stays zero and reads back zero.
- `csrw mhpmevent6` (`0x326`) instead writes `MHPMEVENT_REGW[3]`, so an unrelated CSR write corrupts `mhpmevent3`.
- The top three selectors (`0x33D`–`0x33F`) become unwritable, while `i` = 29, 30, 31 decode to `0x340`–`0x342`, which aren't `mhpmevent` addresses at all.

This also has a second-order effect: `CounterInc[i]` is gated on `MHPMEVENT_REGW[i] != 0`, so a counter configured through the architecturally correct address for its own event selector never actually starts counting.

## Fix

Subtract the three so the write decode lines up with the read decode and with `MHPMEVENTLAST`:

```systemverilog
assign WriteMHPMEVENTM[i] = CSRMWriteM & (CSRAdrM == MHPMEVENTBASE + i - 3);
```

The neighboring `mhpmcounter` write loop does not need the same adjustment — `MHPMCOUNTERBASE` is `mcycle` at index 0, so `+ i` is correct there. `mhpmevent` has no index-0 register, which is what makes its base the odd one out.

## Files changed

- `src/privileged/csrc.sv` (single line)

## Verification

Directed assembly tests on `rv64gc`:
- Write `mhpmevent3`, read it back, and confirm the value matches.
- Write `mhpmevent6` and confirm `mhpmevent3` is left untouched.

Both checks fail on unmodified `main` and pass after the fix.


davidharrishmc (approved):

davidharrishmc:
Thank you for the fix!

PR #1883, opened 2026-09-16
Migrate openhwgroup links to openhwfoundation

davidharrishmc:
The organization renamed, so this points every link at the new name. 313 files, 329 lines.

All but two changed lines are a mechanical `openhwgroup` → `openhwfoundation` inside a URL or repo slug:

- the license header block in the RTL and tests (the bulk of it, 223 `.sv` files and 37 `.S` files)
- the GitHub Actions badge and workflow links in `README.md`
- clone URLs, issue references, and `openhwgroup/cvw Issue #NN` notes in test comments
- the `raw.githubusercontent` fetches in `bin/wally-tool-chain-install.sh`, `docs/docker/Dockerfile.builds`, and the nightly-build line in `README.md`

The two exceptions are the website links in `CONTRIBUTING.md`, which needed real replacements rather than a rename. `openhwfoundation.org` has no `/membership/` or `/working-groups/` page, and the old URLs now redirect to the site root, so a plain domain swap would have produced two 404s:

| old | new |
|---|---|
| `www.openhwgroup.org/membership/` | `openhwfoundation.org/become-a-member/` |
| `www.openhwgroup.org/working-groups/` | `openhwfoundation.org/projects/` |

`/projects/` is the closest page the new site offers; there is no direct successor to the working groups page, so redirect that one elsewhere if you prefer. I also changed "OpenHW Group" to "OpenHW Foundation" in that one sentence, since it reads as the subject of those two links. Other prose uses of the old name are untouched, as are the license and copyright texts.

**Nothing was broken beforehand.** GitHub redirects both web and raw URLs for a renamed organization, so the old links still resolve; this is hygiene, not a fix. I verified the new state rather than assuming it:

- every distinct rewritten URL returns 200, including `raw.githubusercontent.com/openhwfoundation/cvw/...` and both new website pages
- `openhwfoundation/cvw`, `openhwfoundation/cvw-arch-verif` and `openhwfoundation/core-v-verif` all exist
- the diff was checked line by line: every changed line is identical to its predecessor under the rename, except the two `CONTRIBUTING.md` lines above
- `lint-wally` clean on all 7 configs; `arch64i`, `coverage64gc` and `arch32i` pass


MikeOpenHWGroup:
There is no immediate need to make most of these changes.  GitHub will automatically redirect links to https://github.com/openhwgroup/cvw to https://github.com/openhwfoundation/cvw with no action required.

License headers should _not_ be changed:
- The Copyright declared by the OpenHW Group is now owned by the Eclipse Foundation.
- The OpenHW Foundation is a not a legal entity and as such cannot hold a copyright.  (It is a working group of the Eclipse Foundation, which _is_ a legal entity that can hold a copyright.)

The others will require manual edits, but I expect this PR can be greatly reduced in scope.

davidharrishmc:
@MikeOpenHWGroup are you saying we should avoid changing the URL now, or is it fine to do so now instead of later?

I don't see anything related to licenses in the PR.  Am I overlooking something?

MikeOpenHWGroup:
> are you saying we should avoid changing the URL now, or is it fine to do so now instead of later?

Your choice.  You should never _need_ to make this change as GitHub will redirect `openhwgroup` to `openhwfoundation` in perpetuity.  We are confident about this because we still have several documents in GitHub that use pulp-platform/ariane URLs and these are all redirected to openhwfoundation/cva6.

> I don't see anything related to licenses in the PR. Am I overlooking something?

I misinterpreted this line in the Description for this Issue:
```
the license header block in the RTL and tests (the bulk of it, 223 `.sv` files and 37 `.S` files).
```
You are right that these do not include changes to the copyright holder, so that's all good.

Issue #1833, opened 2026-09-04
Review of mmu vs. spec


davidharrishmc:
# Review of `src/mmu` Against the Ratified RISC-V Privileged Specification

**Date:** 2026-09-03
**Scope:** All sixteen files in `src/mmu/` (`mmu.sv`, `hptw.sv`, `pmachecker.sv`, `pmpchecker.sv`, `pmpadrdec.sv`, `adrdec.sv`, `adrdecs.sv`, and `tlb/` with `tlb.sv`, `tlbcontrol.sv`, `tlbcam.sv`, `tlbcamline.sv`, `tlbram.sv`, `tlbramline.sv`, `tlbmixer.sv`, `tlblru.sv`, `vm64check.sv`), plus the LSU align unit, the IFU spill unit, and the config files where the checkers depend on them.
**Reference:** Ratified privileged spec text in `~/repos/riscv-isa-manual/src/priv/` (supervisor.adoc for Sv32/Sv39/Sv48/Sv57 and the translation algorithm, machine.adoc for PMP and PMA, svnapot.adoc, svpbmt.adoc, svadu.adoc), plus Zicbom/Zicboz permission rules and Zicclsm.
**Baseline:** branch `pwm-fixes`, commit `195081020`.

## Summary

The translation, permission, and PMP/PMA logic is largely correct and matches the spec's algorithm step by step. Four definite deviations were found, all in corner cases the arch tests do not exercise, plus one architectural caveat and one observation. No RTL was changed.

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Definite (latent in shipped configs) | `pmpadrdec.sv:37`, `pmpchecker.sv` | PMP matches on the start address only; the "matching entry must match all bytes" rule is not implemented |
| 2 | Definite | `hptw.sv:151-153` | Non-leaf PTEs with reserved bits 63 or 60:54 set are accepted instead of page faulting |
| 3 | Definite | `tlbcontrol.sv:88` | Svnapot N bit accepted on superpage leaf PTEs, which the spec reserves |
| 4 | Definite | `hptw.sv:182-195` | Hardware A/D update is performed on leaf PTEs that must page fault first (misaligned superpage, reserved bits, bad PBMT/NAPOT, S-mode execute of a U page) |
| 5 | Caveat | `hptw.sv` UPDATE_PTE | A/D update is a plain store of the earlier-read PTE, not an atomic read-modify-write; fine for a single hart with no DMA into page tables |
[...]

PR #1889, opened 2026-09-18
mmu: enforce full PMP match across memory operations

Syed-Moeed-Ali:
Partially addresses #1833 (PMP item 1).

**Summary**

The PMP checker currently determines the matching PMP entry using only the starting physical address of a memory operation.

According to the RISC-V privileged specification, the lowest-numbered PMP entry that matches any byte of an access must match all bytes of that access. If a memory operation crosses into another PMP region, the access must fail even if both regions independently allow the requested operation.

**Changes**

**This PR:**

Computes the address of the final byte of each memory operation.
Performs PMP matching on both the starting and ending addresses.
Determines the highest-priority PMP entry for both endpoints.
Raises an access fault when the first and last byte match different PMP entries.
Applies the boundary check independently of the PMP R/W/X permission bits.
Uses the cache-line access size for cache-block operations.
Why this is needed

Previously, an access beginning inside a permitted PMP entry could extend beyond that entry without checking whether the same PMP entry covered the complete operation.

For example, an access crossing between two adjacent readable PMP regions must still fault because no single PMP entry covers the entire access.

This change compares the selected PMP entry for the first and last byte instead of only checking permissions at the ending address.

**Scope**

This PR specifically addresses the PMP full-memory-operation matching issue described in #1833.

Other MMU issues tracked in #1833 are outside the scope of this change.


davidharrishmc:
Thank you for looking at this.  I'm not certain the bug is real.  PMP is subtle.

riscvassertions_wally sets minimum bounds on G when misaligned accesses are possible.  They are intended to avoid any cases where the end needs to be checked.  Also, note that accesses spanning cache line boundaries are split into multiple accesses, each of which should satisfy the PMP rule.

Can you construct any test cases (and configurations) where this fix is truly needed?

This PR adds a significant amount of hardware, so I'm reluctant to add it unless the problem is real.  If the problem is real, let's look for the lowest-cost solution.

```
    assert (P.PMP_G > 0 | P.XLEN == 32 | P.PMP_ENTRIES == 0) else $fatal(1, "RV64 requires PMP_G at least 1 to avoid checking for 8-byte accesses to 4-byte region");
    assert ((P.PMP_G >= $clog2(P.DCACHE_LINELENINBITS/8)-2) | !P.ZICCLSM_SUPPORTED | P.PMP_ENTRIES == 0) else $fatal(1, "Systems that support misaligned data with PMP must have grain size of at least one cache line so accesses that span grains will also cause spills");
    assert ((P.PMP_G >= $clog2(P.ICACHE_LINELENINBITS/8)-2) | !P.ZCA_SUPPORTED | (P.PMP_ENTRIES == 0) | !P.ICACHE_SUPPORTED) else $fatal(1, "Systems that support compressed instructions with PMP must have grain size of at least one cache line so fetches that span grains will also cause spills");
```

Syed-Moeed-Ali:
Closing this. Root-caused the mismatch: our reproducer set PMP boundaries only 4 bytes apart, but PMP_G=4 rounds all boundaries down to 64-byte marks, so both regions collapsed onto the same point and never matched -- meaning PMP was silently disabled for the access, not bypassed via Size. Spike wasn't run with matching grain behavior, hence the false divergence. No real bug here. Thanks for catching this before it landed.

PR #1892, opened 2026-09-18
hptw: retain Misaligned check in LEAF state to prevent spurious A/D write on misaligned superpages

Haiqua-Ghaffar:
## Problem

With Svadu hardware A/D updates enabled, translation through a misaligned
superpage can modify its leaf PTE before raising the page fault. Loads and
instruction fetches with A=0 spuriously set A; stores/AMOs can set A and D,
including setting D when A is already 1.

The walker previously checked alignment only in `Lx_ADR` states. Those checks
became false on entry to `LEAF`, where `HPTWUpdateDA` controls the PTE update.
The subsequent TLB check still raised the page fault, but the PTE had already
changed. Misaligned-superpage validation must precede the conditional A/D
update in the RISC-V address-translation algorithm.

This addresses the misaligned-superpage subcase of #1833, item 4.

## Change

Check `Misaligned` in `LEAF` using the registered `PageType` to select the
appropriate PPN alignment condition. All consumers are gated on `LEAF`, so
the old `Lx_ADR` terms are removed. The change handles RV32 and RV64 without
adding registers or FSM states.

The only file changed in this PR is `src/mmu/hptw.sv`.

## Local validation

Validated locally with a directed Sv39/Svadu suite containing
16 cases covering loads, stores, and instruction fetches at both 2 MiB and
1 GiB page sizes. Aligned controls verify successful accesses and required
hardware A/D updates. Stores cover both A=0,D=0 and A=1,D=0.

Each fault case checks the trap cause, trap value, exception PC, and complete
leaf PTE. The runner requires explicit self-check completion and treats tool
failures and timeouts as errors.

Fresh Verilator builds compare the PR parent (`275695cdb`) with the revised RTL.
The same test ELFs are run independently against Spike and Wally.
The test sources and runner are local validation artifacts and are not included
in this PR.

| Result | Spike | Original Wally | Revised Wally |
| --- | --- | --- | --- |
| Eight misaligned cases | Pass | Exact PTE A/D mismatch after expected page fault | Pass |
| Eight aligned controls | Pass | Pass | Pass |

Verilator lint passes for `rv32gc` and `rv64gc`. Dynamic testing covers RV64
Sv39; it does not cover Sv32/Sv48/Sv57, AMOs, or cache-block operations. No full
regression or ACT generator changes are included. Leaf reserved-bit and
PBMT/NAPOT checks remain outside this PR's scope.


davidharrishmc:
Thank you for this.  I would like to leave this open until the ACTs are enhanced to detect this.

I see you created some directed assembly tests to check this.  Are you familiar with adding tests to the ACT generator?  A Sv generator was just merged, and we need to add tests for all of the MMU items in #1883 that presently are not exercised.  If that is in your skill set and you are looking for things to improve, it would be a useful contribution.

davidharrishmc:
## Claude review

I traced the change through `hptw.sv` rather than relying on the description. **The fix is correct** — I confirmed that every `LEAF`/`PageType` pairing in the patch matches its existing `Lx_ADR` pairing (checked for Sv32/Sv39/Sv48/Sv57), that `PTE`/`CurrentPPN` are still stable in `LEAF` (`PRegEn` at :153 zeroes `PTE` only on the edge *leaving* `LEAF`), and that the replayed access does fault without livelocking (`tlb.sv:107-111` → `tlbcontrol.sv:123`, with `UpdateDA` gated by `~TLBPageFault` at `tlbcontrol.sv:120`).

Two comments on the patch as written.

### 1. The `WalkerState == Lx_ADR` terms are now dead logic — replace them rather than OR onto them

`Misaligned` can only ever have an effect while `WalkerState == LEAF`:

- `Misaligned` feeds only `OtherPageFault` (:210)
- `OtherPageFault` feeds only `HPTWUpdateDA` (:216)
- `HPTWUpdateDA` is consumed at :219 (`UpdatePTE = (WalkerState == LEAF) & HPTWUpdateDA`), :235/:236 (`DTLBWriteM`/`ITLBWriteF`, both gated on `WalkerState == LEAF`), and :330 (the `LEAF:` arm of the FSM)

All four consumers are gated on `LEAF`, so an assertion of `Misaligned` in an `Lx_ADR` state cannot reach anything. That is the PR's own premise restated: if the old check had had an effect, there would be no bug to fix.

Keeping both halves therefore doubles the width of an already dense expression and implies the `Lx_ADR` check still does something. Suggest replacing outright:

```systemverilog
// RV64
assign Misaligned = (WalkerState == LEAF) &
                    ((P.SV57_SUPPORTED & (PageType == 3'b100) & PetapageMisaligned) |
                     (P.SV48_SUPPORTED & (PageType == 3'b011) & TerapageMisaligned) |
                                        ((PageType == 3'b010) & GigapageMisaligned) |
                                        ((PageType == 3'b001) & MegapageMisaligned));

// RV32
assign Misaligned = (WalkerState == LEAF) & (PageType == 3'b001) & MegapageMisaligned;
```

This also keeps the lines near the file's usual width; the current RV64 lines run to roughly 150 characters.

If you believe the Lx_ADR terms are load-bearing after all, could you say which consumer sees them? That would be worth a comment in the code either way.

2. The bug is broader than store/AMO

systemverilog
assign SetDirty    = ~Dirty & DTLBWalk & (WriteAccess | CMOpM[3]);              // :195
assign HPTWUpdateDA = ValidLeafPTE & (~Accessed | SetDirty) & ENVCFG_ADUE & ~OtherPageFault;  // :216

The ~Accessed term carries no DTLBWalk or WriteAccess dependence, so a load or an instruction fetch to a misaligned superpage whose PTE has A=0 also causes the spurious UPDATE_PTE and the A-bit write — not just stores and AMOs. (For an ITLB walk DTLBWalk=0, so SetDirty=0 and ~Accessed is the only term that can fire.)

The patch fixes all three cases, which is good, but the problem statement and the three listed tests only cover the store case. Worth widening the description, and adding a load and an instruction-fetch case so the other two paths are actually exercised.

abdulrafay7038:
Thanks, David. I'm the part of the same team as Haiqua.
I see your point that this is a broader issue than the specific case covered by the current fix. I’ll go back through the affected RTL path and update the patch to address the underlying condition more generally rather than only handling the original reproducer. I’ll also add directed regression coverage for the broader cases so we can verify the fix across the affected scenarios. For now I’ll keep this PR focused on the RTL fix and its regression coverage, and update the PR once I’ve validated the broader solution.

davidharrishmc:
@abdulrafay7038 thank you.  Does your team have capabilities to enhance the riscv-arch-test Sv suite to have detected all these CVW issues?

abdulrafay7038:
Thanks, David. At the moment, we don’t have much experience working directly with the riscv-arch-test Sv suite, and our current focus is on the WallyGuard project itself, developing and evaluating the agentic bug-discovery, reproduction, and RTL-fix workflow.

We’d definitely be interested in looking into improving the architecture tests in the future, especially once we have more experience with the suite and can better understand how to generalize the issues we’re finding into reusable tests. For now, though, our priority is to continue validating and fixing the broader CVW bugs we uncover through WallyGuard.

abdulrafay7038:
Thanks David. I would like to know whether it is a paid opportunity.

davidharrishmc:
Sounds good.  We are on a short time frame, so we will improve the arch tests.

davidharrishmc:
Glad you are developing WallyGuard and finding and fixing bugs.  This is a valuable contribution.

abdulrafay7038:
Thanks for tracing the consumers. We agree: the `Lx_ADR` terms have no effect because every consumer of `HPTWUpdateDA` is gated on `LEAF`. The revision replaces them with a single `WalkerState == LEAF` gate and the appropriate `PageType` check for RV32 and RV64.

The description now covers loads and instruction fetches as well as stores/AMOs. The existing fix already blocks all of those update paths; the RTL revision simplifies that fix.

We validated the revision locally with a directed Sv39/Svadu suite covering loads, stores, and instruction fetches on both 2 MiB and 1 GiB pages. Each has an aligned control with A/D initially clear. Stores also test A=1,D=0, so `SetDirty` is exercised independently of `~Accessed`.

The 16 test ELFs run independently on Spike and Wally. The fault cases check the cause, trap value, exception PC, and complete leaf PTE. They report expected/actual values through `selfcheck_record`, so the regression identifies the PTE mutation directly rather than inferring it from a watchdog timeout.

Validation against the PR's parent, `275695cdb`, and a fresh build of the revision:

- Spike passes all 16 cases.
- Before the fix, all eight misaligned cases reach the expected page fault and fail specifically on the PTE comparison. Loads/fetches set A; stores set A/D or D. All eight aligned controls pass.
- With the revision, all 16 cases pass on Wally, including the controls that require hardware A/D updates.
- Verilator RTL lint passes for `rv32gc` and `rv64gc`.

This update changes only `src/mmu/hptw.sv`. The directed tests and runner remain local validation artifacts and are not included in the PR. Dynamic coverage here is Sv39; Sv32/Sv48/Sv57 and AMOs are not claimed as dynamically tested.

Issue #1811, opened 2026-08-05
Remove RISCOF dependency


jordancarlin:
RISCOF depends on `riscv-isac`, which in turn depends on `python-constraint`. `python-constraint` was last released in 2018 and is distributed in a format that is no longer compliant with how Python PyPI packages are distributed. That latest version of uv (0.12.0) dropped support for the legacy installation formats, so it is no longer able to install/run Wally Python scripts. We need to drop the legacy packages and move Wally fully over to ACT4. This is blocking https://github.com/riscv/riscv-arch-test/pull/2062.

PR #1895, opened 2026-09-18
Retire riscof: self-checking riscv-arch-test runs, with code and functional coverage restored

davidharrishmc:
## Summary

Wally's tests are now entirely self-checking ELFs that report through `tohost`; the signature-checking testbench path, `tests.vh`, riscof, and wally-riscv-arch-test are gone. With the test flow rebuilt on riscv-arch-test (ACT), code coverage and SystemVerilog functional coverage are both working again.

17 commits, linear on `main`.

### Self-checking tests replace riscof

- **wsim** takes any mix of ELF files and directories (`*.elf` found recursively) and runs them back-to-back in one simulation. It pre-generates memfiles in parallel, resolves `tohost`/`begin_signature` with `nm`, and passes the list via `+ElfList`. `--test {buildroot,fpga,coremark,embench}` selects special modes; `--name` names log/ucdb files.
- **testbench** runs each ELF in turn, captures the `tohost` store (1 = pass, `(code<<1)|1` = fail, no store = fail), and prints a single `SUCCESS`/`FAIL` summary. Signature and self-check-record checking, `tests.vh`, and label-map parsing are removed; label maps are only built for the `functionName` debug tracer.
- **regression-wally** builds ACT ELFs for rv32imc/rv32gc/rv64gc from cvw-owned configs in `config/<cfg>/act/`, runs one simulation per ACT leaf directory in parallel, keeps compiled designs between runs, runs periph tests as explicit groups, and regenerates ELFs when a config changes. Standard regression: **~2.5 min**.
- **Submodules**: the act3 `riscv-arch-test` and `riscv-arch-test-cvw` submodules are replaced by one `addins/riscv-arch-test` on `act4`; `cvw-arch-verif` references removed.
- **bin/actconfig-sync** (run by `lint-wally`) checks each configuration's UDB yaml and `sail.json` against the `*_SUPPORTED` parameters in `config.vh`, including that a memory region only advertises compare-and-swap atomics where Zacas is implemented. It caught Zcmop, Zimop, Zabha and Zawrs still listed for configurations that do not implement them, `AMOCASD`/`AMOCASQ` regions without Zacas, and `COUNTERS = 0` on rv32e/rv32i/rv64i, which is below the three machine-mode counters every hart must implement.
- periph/coverage test libraries report failures through `tohost`; `extractFunctionRadix` dropped from test Makefiles; riscof removed from the Docker images.

### Code coverage

`--ccov` was dead: two exclusions in `coverage-exclusions-rv64gc.do` named signals that no longer exist, `GetLineNum` raised an error, Questa's `onerror` aborted the do-file, and `coverage save` never ran, so no `.ucdb` was written. With that unblocked and the stale exclusions repaired, rv64gc goes from **98.30% to 99.3%** (priv 91.19 → 99.80%, ifu 99.50 → 99.86%; ieu/ebu/fpu/hzu at 100%), through new tests in `tests/coverage` for the performance counters, machine-mode supervisor interrupts, the supervisor WFI trap, a reserved `c.ldsp`, and the cache-management PMP checks with PMP unenforced, plus justified exclusions where the RTL is genuinely unreachable.

Coverage is also reusable: a rerun with nothing changed drops from **11 minutes to 32 seconds**, since a test's `.ucdb` is kept when it is newer than both its ELF and every design source (the exclusions file included, because exclusions are applied when each `.ucdb` is saved).

The PWM peripheral test could not run in lockstep because `imperas.ic` had no PWM region and the testbench did not mark it volatile; with both fixed the waiver is gone and `pwmdec` is covered.

### Functional coverage

`--fcov` was inert. It now compiles the ACT covergroups against the cvw testbench, deriving the covergroup defines from the ELF directories actually built for the configuration, and reports **99.50% covergroup coverage over 540 covergroup types** (99.66% of 1,734,200 bins) for rv64gc. The remaining gaps are the Sv and PMP groups, which are equally uncovered in ACT's own reference runs.

### Build speed

Verilator had no `-j` and always built with `--trace`; the generated C++ dominated the build. Building in parallel and tracing only when a VCD is wanted takes a clean build of a small configuration from **61.7 s to 11.5 s**.

## Folded in from #1901: riscof dependencies and the installation workflow

- `pyproject.toml` drops riscof, riscv-config and riscv-isac, and the lock is regenerated: **137 packages to 40**. Gone with them are `python-constraint` 1.4.0, the legacy archive uv 0.12 cannot unpack (the blocker in #1811), riscv-isac's documentation and publishing stack installed as runtime dependencies, and `pyyaml` 5.2 (CVE-2020-1747). `uv-install.sh` moves to uv 0.12.13, the version riscv/riscv-arch-test#2062 moves ACT to.
- `install.yml` no longer runs `make riscof` and `wsim rv32i arch32i`, both of which this PR removes; it builds the self-checking peripheral tests and runs `wsim rv32imc tests/periph/rv32/WALLY-periph-m-01.elf --sim verilator`, which exercises the installed compiler, Verilator and the whole `wsim` flow without the reference-model toolchain an ACT build wants.
- `bin/fparchtest.sh` (referenced nowhere) is deleted, and the docker regression knob `BUILD_RISCOF` is renamed `BUILD_ACT`.

These merge together so the installation workflow works at every commit on `main`.

## Submodule

riscv-arch-test [#2420](https://github.com/riscv/riscv-arch-test/pull/2420) and [#2423](https://github.com/riscv/riscv-arch-test/pull/2423), which let the covergroups compile for a DUT without the vector extension, merged on September 19, so the submodule now points at the `act4` head (`fce1220e8`) rather than a pull-request commit.

## Not yet done (follow-up PRs)

- Derivative configurations still name old `tests.vh` suites, so `--nightly` is broken until they are converted to ACT directories by extension and `derivgen.pl` emits `config/deriv/<cfg>/act/`.
- Per-configuration `coverage.svh` management.
- rv32e, rv32i and rv64i have no ACT configuration yet, so they drop out of the standard regression (they ran `arch32e`, `arch32i` and `arch64i` before).
- `--ccov` reuse keys on file times alone: a test that failed still saves its `.ucdb`, so its failure is hidden on the next run until a source changes. Reuse should require that the previous run passed.

## Test plan

- `regression-wally`: rv32imc/rv32gc/rv64gc ACT + periph + buildroot-short + lint, green in ~2.5 min.
- `regression-wally --ccov`: 1324 ELFs in lockstep, green; merged report as above.
- `regression-wally --fcov`: 98 covergroup runs, green, 99.50%.
- Mixed ELF group (coverage + periph + ACT) on rv64gc with Verilator and Questa; known-bad `WALLY-uart-timeout-01` correctly reports `tohost = 5`.
- `actconfig-sync --check` clean; `make -n all`, `make -n -C tests/fp`, `make -n -C sim` parse.


jordancarlin:
I'll review this once I'm caught up on the urgent ACT backlog. Excited to get this in!

PR #1896, opened 2026-09-18
Drop obsolete Debian 11 through installation, and fix Mold guard

davidharrishmc:
Two independent installation-script fixes. The sourceware.org retry this PR originally carried is gone in favor of #1898, which fixes the rate limiting at the root by bumping `riscv-gnu-toolchain` to a commit whose submodules come from GitHub mirrors.

**Debian 11.** Bullseye is past LTS end of life and its package pool has moved to `archive.debian.org`, so an install on Debian 11 now fails in apt before any of this code can run:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/g/git/git_2.30.2-1+deb11u5_amd64.deb  404  Not Found
```

The minimum supported Debian version becomes 12, and the workarounds that existed only for 11 are gone: the python3.9 package selection, the mold binary download, and the Whisper skip. README compatibility notes updated to match. #1897 drops the matching CI job.

Behavior is unchanged for every still-supported distro: on Debian, `UBUNTU_VERSION` is unset, so `(( UBUNTU_VERSION != 20 ))` is true exactly where the old two-clause test was.

**Mold guard.** `SUSE_VERSION` is exported only on SUSE, so on every other distro the guard in `verilator-install.sh` evaluated `(( SUSE_VERSION < 160 ))` as `0 < 160` and downloaded the mold binary into `$RISCV` even though `wally-package-install.sh` had just installed mold from the package manager. On `ubuntu-24.04-arm` it fetched an x86_64 binary onto an aarch64 machine. Giving the comparison a lower bound restricts it to SUSE, so the download now happens exactly where no mold package is installed:

| distro | mold package | download |
|---|---|---|
| Ubuntu 20.04 | no | yes |
| Ubuntu 22.04 / 24.04 / 26.04 | yes | no |
| Debian 12 / 13 | yes | no |
| Red Hat / Rocky / AlmaLinux 8, 9, 10 | yes | no |
| SUSE 15.6 | no | yes |
| SUSE 16.0 | yes | no |

**Verification.** The mold guard was checked against the table above with the relevant variables set the way `wally-environment-check.sh` sets them. `bash -n` is clean on every touched script, and shellcheck reports only pre-existing warnings.

Not changed here: the mold tarball URL is hardcoded to `x86_64`. With the guard fixed that only matters for Ubuntu 20.04 or SUSE 15.6 on ARM, neither of which is in CI, but `$(uname -m)` would cover it since mold publishes aarch64 assets.


jordancarlin:
Newer versions of the `riscv-gnu-toolchain` repo have migrated to more reliable mirrors and will be a more robust solution for this.

jordancarlin (approved):

PR #1897, opened 2026-09-18
Drop the Debian 11 installation job

davidharrishmc:
DH: Merge before 1896.

Workflow-only half of #1896, split out so it can merge on its own.

**Debian 11.** Bullseye is past LTS end of life and its package pool has moved to `archive.debian.org`, so the job now fails in apt before the installation script even starts:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/g/git/git_2.30.2-1+deb11u5_amd64.deb  404  Not Found
```

It passed on the September 2 nightly and has failed every run since September 9. Pointing it at `archive.debian.org` was the alternative, but keeping an archived distro green costs more than it proves.


jordancarlin:
See my comment on #1896. This is not the best way to solve this and will make CI take upwards of 6 hours.

davidharrishmc:
I merged this without going through CI because the obsolete Debian was blocking other PRs from getting through CI.

jordancarlin:
If we are dropping Debian 11 since it is EOL, is it time to drop Ubuntu 20.04 too? It went EOL in May 2025.

davidharrishmc:
I dropped Debian 11 because it was unreachable and breaking CVW CI.  
Is Ubuntu 20.04 breaking anything or keeping us on bad tool versions?

jordancarlin:
I don’t recall. I think the installation scripts have a number of hacks to work around Ubuntu 20.04 limitations, but as far as I know, those all continue to work, so maybe no reason to touch it.

PR #1901, opened 2026-09-19
Remove the riscof, riscv-config and riscv-isac dependencies and move to a current uv

### LINKED ISSUE #1811: Remove RISCOF dependency

davidharrishmc:
Stacked on #1895, which retired the riscof signature tests; the diff is the last three commits. Closes #1811 and unblocks riscv/riscv-arch-test#2062 (the uv 0.12 bump), which #1811 names as blocked on this.

### Why

Nothing in cvw imports or runs riscof, riscv-config or riscv-isac after #1895, but `pyproject.toml` still pulled all three from pinned git commits, and through them 97 more packages. One of those is `python-constraint` 1.4.0, last released in 2018 and shipped as a legacy archive that uv 0.12 can no longer unpack. Reproduced with uv 0.12.13 against the current lock, no cache:

```
  × Failed to download and build `python-constraint==1.4.0`
  ├─▶ Failed to extract archive: python-constraint==1.4.0
hint: `python-constraint` (v1.4.0) was included because `cvw` (v0.9.0) depends on `riscv-isac` (v0.18.0) which depends on `python-constraint`
```

So a current uv cannot install cvw's Python environment at all — that is the blocker #1811 describes.

### What changes

**Dependencies.** The three are dropped and the lock regenerated; `uv lock --upgrade` also refreshes the resolutions the removed pins had frozen. The lock goes from **137 packages to 40**. Removed along with them:

- `python-constraint` 1.4.0 — the uv 0.12 blocker
- riscv-isac's documentation and publishing stack, installed as *runtime* dependencies: `sphinx` + eight `sphinxcontrib-*`, `twine`, `keyring`, `cryptography`, `requests`, `urllib3`, `pytest`, `m2r2`, `mistune`, `doc8`
- the deprecated `gitdb2` / `smmap2` shims, `oyaml`, `pbr`, `colorlog`, `ruamel-yaml`, `cerberus`, `pytablewriter`
- `pyyaml` **5.2**, which the old pins had held below 5.3.1 (CVE-2020-1747, `full_load` code execution); now 6.0.3

What remains is what cvw's own scripts use: `adjustText`, `lief`, `Markdown`, `matplotlib`, `pre-commit`, `scikit-learn`, `scipy`.

**Dead riscof-era code.** `bin/fparchtest.sh` (an untested driver for `riscv-ctg`/`riscv-isac`, referenced nowhere) is deleted; the docker regression script's `BUILD_RISCOF` switch already built the ACT tests and is renamed `BUILD_ACT` (Makefile, README, `run_regression.sh`); a `regression-wally` comment no longer refers to a riscof config file. `errata.md` keeps its historical mentions.

**Installation workflow.** It still ran `make riscof` — a target #1895 removed — and then `wsim rv32i arch32i`, a riscof-era test name `wsim` no longer resolves. The paths filter never triggered the workflow on #1895, so this break was latent until the weekly run. It now builds the self-checking peripheral tests and runs `wsim rv32imc tests/periph/rv32/WALLY-periph-m-01.elf --sim verilator`, which exercises the installed compiler, Verilator and the whole `wsim` flow without needing the reference-model toolchain an ACT build wants (Ruby/UDB via mise is installed but never activated in that job). Passes locally. Building ACT inside the install job is a reasonable follow-up.

**uv pin.** `uv-install.sh` moves from 0.11.0 to **0.12.13**, the version #2062 moves riscv-arch-test to, so the two agree.

### Verification

- `uv lock` written with uv 0.11.33 (the version installed here, and ACT's current `.mise.toml` pin), so it stays readable by existing installs.
- uv 0.11.33: `uv sync --frozen`, import of every remaining direct dependency, `wsim --help`, `regression-wally --help` — all OK.
- uv 0.12.13 (scratch install): `uv lock --check` clean; `uv sync --frozen --no-cache` installs 30 packages; imports OK. The same command against the *old* lock fails with the `python-constraint` error above.
- Not yet exercised: a full `regression-wally` run from the reduced environment. `wsim` and `regression-wally` start and parse arguments under it; the scripts import only the standard library plus the packages that remain.
- `install.yml` parses; the smoke test runs green locally on Verilator.


davidharrishmc:
Folded into #1895 as its last four commits (fc224880b, 8d491cc74, bb80a905f, plus the submodule pointer move to act4), so the installation workflow keeps working at every commit on main rather than breaking between two merges. Closing in favour of #1895.

